PR #2, opened 2015-10-17
First prototype of php-scoper

matthieuauger:
This is the beginning of php-scoper implementation (cf https://github.com/webmozart/php-scoper/issues/1). Here is the checklist of what is done, and what is remaining:

- [x] Manage single files
- [x] Manage multiple files (should be good, not tested yet)
- [x] Manage directories
- [x] "Namespace" prefixing
- [x] "Use namespace" prefixing
- [x] "FQ namespace" prefixing
- [x] Manage single files
- [ ] Keep coding style of input file
- [x] Output management (what do we display ?)
- [x] Error management
- [x] Tests

@webmozart : Can I have your opinion of this implementation ?

How I see the next steps:

- Find a way to replace an expression "on the fly". Currently all the tree is updated, then dumped into the initial file. I lose all the initial coding style
- Do you have an idea of what should be displayed to the user when he runs the command ?
- I named my class "Scoper" with a "Scope" function. Maybe I should call it "Prefixer" with a "Prefix" function instead. I don't know, but I can change the names if it is more appropriate 

All feedback is really welcome


src/Scoper.php
```diff
@@ -0,0 +1,56 @@
...
+        $traverser->addVisitor(new UseNamespaceScoperNodeVisitor($prefix));
+        $traverser->addVisitor(new FullyQualifiedNamespaceUseScoperNodeVisitor($prefix));
+
+        //TODO Manage errors
+        $statements = $this->parser->parse($content);
+        $statements = $traverser->traverse($statements);
+
+        $prettyPrinter = new Standard();
+
+        return $prettyPrinter->prettyPrintFile($statements)."\n";
```
matthieuauger: The prettyPrinter does not keep coding style, nor newlines at the end of files. This should be improved with "on-the-fly" replacement

matthieuauger:
I don't really understand why appveyor is complaining!

webmozart:
Awesome, thanks! :) Will review tomorrow.

cc @stof @sebastianbergmann @naderman @seldaek

matthieuauger:
Do you have a real-life example where this tool could be used ? This would be the best check

webmozart:
@matthieuauger You can checkout basically any code repository, apply the tool to both the source code and the Composer vendors and see whether the code works (e.g. run the tests). The only prerequisite is that the code doesn't use dynamic classes, i.e. isn't building class names through strings and then creating them this way. For a start you could try https://github.com/puli/cli.

src/Handler/AddPrefixCommandHandler.php
```diff
@@ -22,6 +26,30 @@
...
+     * @var Finder
+     */
+    private $finder;
+
+    /**
+     * @var Scoper
+     */
+    private $scoper;
+
+    public function __construct(Filesystem $filesystem, Finder $finder)
```
webmozart: `Filesystem` and `Finder` are utility classes that are not meant to be replaced. I think you shouldn't inject them, but simply construct them internally.

src/Scoper.php
```diff
@@ -0,0 +1,54 @@
...
+        $this->parser = $parser;
+    }
+
+    /**
+     * @param $content
+     * @param $prefix
+     *
+     * @return string
+     */
+    public function scope($content, $prefix)
```
webmozart: IMO this should be called `addNamespacePrefix()` for more clarity.

src/Scoper.php
webmozart: I think that instead of changing `$content`, you should collect all the offsets where you want to insert the namespace prefix and them insert them into the original content in one go. That way you can maintain the original file format.

See also https://github.com/webmozart/php-scoper/issues/1#issuecomment-148896044.

src/Handler/AddPrefixCommandHandler.php
matthieuauger: Yes, I thought it would be easier to test the CommandHandler by mocking the dependencies, but the tests are using the real Filesystem and Finder, so we can construct them internally

src/Scoper.php
matthieuauger: Absolutely !

src/Scoper.php
matthieuauger: Indeed I have the offsets, I can stack the changes. I'll work on it

matthieuauger:
@webmozart : Thanks for your feedback ! I tried to run the scoper on puli, and the prefixing of all the code has been successful.

But I ran into some issues:
- Composer has some checks based on [string comparison](https://github.com/composer/composer/blob/15face5432d7b7334db6ac69fac0190971cafa6e/tests/Composer/Test/Autoload/Fixtures/autoload_real_files_by_dependency.php#L11). As a workaround, I ran `composer dump-autoload` again after the scoper.
- All the composer autoloading is incorrect after the scoper has run, I think it's because the "autoload" section in composer.json of each vendor does not contain the prefix that we add.
- A lot of classes in PHPUnit don't have a namespace ([TestCase](https://github.com/sebastianbergmann/phpunit/blob/master/src/Framework/TestCase.php) for example). I'll check again but actually no replacement is made in thoses files

First, I'll fix update this PR to take in consideration your comments and add the "on-the-fly" replacement. But I don't know how yet to address these issues above, do you have any ideas ?

Thanks

sebastianbergmann:
@matthieuauger Types that are not declared in a namespace should be moved into a namespace.

webmozart:
@matthieuauger I agree with @sebastianbergmann. I would solve the other issues you mentioned in separate PRs - I created separate tickets for them:

* #3: Add --diff option
* #4: Add --fix-strings and --fix-global-strings options
* #5: Add --fix-composer option

In practice, the workflow would be something like this:

```
$ php-scoper add-prefix --fix-php --fix-strings --fix-composer MyPrefix\\ src vendor
$ composer dump-autoload
$ php-scoper add-prefix --fix-php --fix-strings MyPrefix\\ vendor/autoload.php vendor/composer
$ php-scoper add-prefix --fix-global-strings --diff MyPrefix\\ src vendor > global-strings.patch

review global-strings.patch and eliminate false positives

$ cat global-strings.patch | patch -p0
```

In this example, the `--fix-php` option controls what you are implementing right now. I think you should add this option so that the scoper can be run without this feature.

I'm not sure about the `--fix-` prefix of the options, so please suggest something better if you can.

matthieuauger:
@sebastianbergmann All right, I can propose a PR for that. The scoper can help me find all those classes.

@webmozart : I'll take a look on these issues. I'm not sure either for the --fix option, as we are not really "fixing" something, but I don't have any other suggestions at the moment, I'll think about it and we can still develop the feature.

Thanks

src/Scoper.php
nikic: This is the class I use for doing these kinds of changes: https://github.com/nikic/TypeUtil/blob/master/src/MutableString.php

webmozart:
@matthieuauger What's the status of your PR?

matthieuauger:
@webmozart : There is still a lot of work remaining I think, at least the missing checks mentioned above (keep coding style, output management, error management) + Adding the namespace to non-namespaced classes (I don't know yet how to effectively do that) + the 4 issues you created.

I can finish the checks above but would love some help for the remaining issues.

webmozart:
Ok. I think the best way to proceed is to do what's necessary to finish this PR, merge it and do the rest in follow-up PRs. Can you do that? Could you also create tickets for the remaining work? This way it's easier to delegate this to other people.

matthieuauger:
@webmozart : I also think this is the best strategy, I'll try to finish it during the week

webmozart:
Awesome, thanks!

matthieuauger:
Hi @webmozart,

I added some output when scoping the files and error management for parsing exceptions. You can check https://github.com/matthieuauger/php-scoper/commit/73b1058d96db45c9133705a30d0c0801e27bc0c4 and https://github.com/webmozart/php-scoper/commit/296e5db098f675269feb40fdcfea7d22dc16326c

There is still #7 to do, but everything here is tested and already usable without it, I think this could be merged without it.

I still have a issue when running tests on Windows (via appVeyor), I don't know if the FileSystem component is able to remove non-empty directories and don't have a Windows myself to check locally what the problem is. Do you have any ideas ?

Thank you

tests/Handler/AddPrefixCommandHandlerTest.php
```diff
@@ -82,12 +82,61 @@ protected function tearDown()
 
     public function testAddPrefixToDirectory()
     {
-        chdir(dirname($this->tempDir));
+        chdir($this->tempDir);
```
webmozart: This is the reason why the tests fail on AppVeyor. On Windows, you cannot remove directories that are or contain the current working directory you should:

* store the current working directory in `setUp()` (with `getcwd()`)
* restore the working directory in `tearDown()` before you delete the temporary directory

matthieuauger: Good catch, the tests are now passing !

matthieuauger:
@webmozart : I fixed the tests, and all the remaining work is listed in the [issues page](https://github.com/webmozart/php-scoper/issues).

Do you have anymore feedback ?

Thanks

tests/Handler/AddPrefixCommandHandlerTest.php
```diff
@@ -68,26 +73,77 @@ protected function setUp()
...
         $filesystem->remove($this->tempDir);
     }
 
     public function testAddPrefixToDirectory()
     {
-        chdir(dirname($this->tempDir));
+        chdir($this->tempDir);
+
+        $args = self::$command->parseArgs(
+            new ArgvArgs(
```
webmozart: You can simplify this (and the other tests) by using `StringArgs()`:

```php
new StringArgs('add-prefix MyPrefix\\ dir/dir')
```

webmozart:
Apart from my last comment, this looks good. :)

tests/Handler/AddPrefixCommandHandlerTest.php
matthieuauger: Are you sure ? I'm getting a `Webmozart\Console\Api\Args\CannotParseArgsException: Not enough arguments.` when replacing the ArgvArgs with StringArgs

webmozart: My bad, this should have been:

```php
new StringArgs('MyPrefix\\ dir/dir')
```

matthieuauger:
@webmozart : Time flies... I used the StringArgs class instead of ArgvArgs as you suggested, thanks for the tip

webmozart:
Great, thanks! :)

PR #11, opened 2015-12-03
Keep coding style when scoping files

### LINKED ISSUE #7: Keep coding style during prefixing process
Currently, the coding style is not kept when files are processed by the scoper. For exemple

```php
<?php

namespace MyNamespace;

use MyOtherNamespace\MyClass;

class MyClass
{
}
```

is modified like this

```php
<?php

namespace MyNamespace;

use MyOtherNamespace\MyClass;
class MyClass
{
}
```

matthieuauger:
Hi,

This PR fixes #7 

This code works well only because the MutableString is a reference and is shared between all Visitors. I'm not sure this is a good design, any feedback welcome.

Thanks

### DIFF
--- a/src/NodeVisitor/NamespaceScoperNodeVisitor.php
+++ b/src/NodeVisitor/NamespaceScoperNodeVisitor.php
@@ -6,23 +6,35 @@ use PhpParser\Node;
 use PhpParser\Node\Name;
 use PhpParser\Node\Stmt\Namespace_;
 use PhpParser\NodeVisitorAbstract;
+use Webmozart\PhpScoper\Util\MutableString;
 
 class NamespaceScoperNodeVisitor extends NodeVisitorAbstract
 {
+    /**
+     * @var MutableString
+     */
+    private $content;
+
     /**
      * @var string
      */
     private $prefix;
 
-    public function __construct($prefix)
+    public function __construct(MutableString $content, $prefix)
     {
+        $this->content = $content;
         $this->prefix = $prefix;
     }
 
     public function enterNode(Node $node)
     {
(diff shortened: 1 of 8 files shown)


matthieuauger:
@webmozart : Any feedback on this PR ?

Thanks

src/NodeVisitor/NamespaceScoperNodeVisitor.php
```diff
@@ -6,23 +6,35 @@
...
     {
+        $this->content = $content;
         $this->prefix = $prefix;
     }
 
     public function enterNode(Node $node)
     {
         if ($node instanceof Namespace_ && null !== $node->name) {
-            $node->name = Name::concat($this->prefix, $node->name);
+            $prefix = (0 === strpos($node->name, '\\')) ? $this->prefix : $this->prefix.'\\';
```
webmozart: What exactly does this condition do?

webmozart:
Hi @matthieuauger, thanks for adding this :) looks good! 

Can you add a tests for the individual classes?

webmozart:
@matthieuauger any news?

webmozart:
ping?

webmozart:
ping @matthieuauger

afurculita:
@matthieuauger any news on this?

padraic:
@afurculita Working on an update to this in a personal branch which I can PR when ready. Seems mostly tests and a few small bits and pieces (e.g. licence/attribute for new class(es)).

theofidry:
@padraic and I don't think this is worth pursuing: it adds non significant complexity for little value. We don't think the consumers will care of the CS of the prefixed code.

PR #16, opened 2017-05-21
Update dependencies and tests

padraic:
Does what it says on the tin ;). Allows composer to install correctly; gets tests to passing state.

A few test updates add additional spaces to fixtures used as expected output. I'll leave it as separate exercise to see where those are added, and if they can be conditionally omitted.

Note: Updated Appveyor config and extended separately with #17. The failure at present is just choco needing Windows Update to run for its missing deps.

### DIFF
--- a/tests/Handler/AddPrefixCommandHandlerTest.php
+++ b/tests/Handler/AddPrefixCommandHandlerTest.php
@@ -11,7 +11,7 @@
 
 namespace Webmozart\PhpScoper\Tests\Handler;
 
-use PHPUnit_Framework_TestCase;
+use PHPUnit\Framework\TestCase;
 use Symfony\Component\Filesystem\Filesystem;
 use Webmozart\Console\Api\Command\Command;
 use Webmozart\Console\Args\StringArgs;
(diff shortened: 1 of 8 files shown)


tests/Handler/AddPrefixCommandHandlerTest.php
```diff
@@ -94,9 +94,9 @@ public function testAddPrefixToDirectory()
         $args = self::$command->parseArgs(new StringArgs('MyPrefix\\\\ dir'.DIRECTORY_SEPARATOR.'dir'));
 
         $expected = <<<EOF
-Scoping $this->tempDir/dir/dir/MyClass.php. . . Success
 Scoping $this->tempDir/dir/dir/MySecondClass.php. . . Success
 Scoping $this->tempDir/dir/dir/MyThirdClass.php. . . Success
+Scoping $this->tempDir/dir/dir/MyClass.php. . . Success
```
padraic: Order of MyClass line is acting up. Passes on 5.6-7.1 with/without xdebug locally. Fails on all builds with Travis.

PR #17, opened 2017-05-21
Update Appveyor config; test for PHP 5.6, 7.0, 7.1

padraic:
- Extends PHP versions to test against.
- Enables Windows Update (choco depends on updates some of the time).
- Not sure about x86 vs x64, so assumed a modern system and switched.
- Enabled caching to speed up builds.

Note: Build will fail pending #16.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,37 +1,63 @@
 build: false
-platform: x86
+platform:
+  - x64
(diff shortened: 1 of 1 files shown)


theofidry:
I'm aware the AppVeyor tests are still failing, will take a look at it later

PR #19, opened 2017-05-27
Force delete the test temp directory on Windows

padraic:
Currently creating an error on Appveyor build using Symfony Filesystem. This instead executes `rd /s` command to delete the entire tree including files.

### DIFF
--- a/tests/Handler/AddPrefixCommandHandlerTest.php
+++ b/tests/Handler/AddPrefixCommandHandlerTest.php
@@ -93,7 +93,7 @@ class AddPrefixCommandHandlerTest extends TestCase
 
     public function testAddPrefixToDirectory()
     {
-        chdir($this->tempDir); exit;
+        chdir($this->tempDir);
 
         $args = self::$command->parseArgs(new StringArgs('MyPrefix\\\\ dir'.DIRECTORY_SEPARATOR.'dir'));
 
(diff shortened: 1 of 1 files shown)


padraic:
I'm obviously a little too distracted to be writing code this evening ;). Stopping now before I make typos 3 through 99.

tests/Handler/AddPrefixCommandHandlerTest.php
```diff
@@ -84,7 +84,11 @@ protected function tearDown()
     {
         chdir($this->workingDirectory);
         $filesystem = new Filesystem();
-        $filesystem->remove($this->tempDir);
+        if (!defined('PHP_WINDOWS_VERSION_BUILD')) {
+            $filesystem->remove($this->tempDir);
+        } else {
+            exec(sprintf("rd /s /q %s", escapeshellarg($this->tempDir)));
```
theofidry: why is this workaround needed? I would expect the filesystem to take care of such platform specific issues

padraic: I'm not sure I'd call it a work around, it's a one line command which does the same thing as spending tens of lines writing something with a directory iterator. The Filesystem component is up to 3.2 and doesn't seem to handle this case.

theofidry: I'm gonna check but IMO it is something it should handle. In any case it's not gonna be fixed immediately so let's go with it for now

theofidry:
Closed via #29

PR #20, opened 2017-05-27
Fix StyleCI misconfiguration

padraic:
"The provided fixer 'empty_return' cannot be disabled unless it was already enabled by your preset."


padraic:
Note: Enabled StyleCI access - forgot it while setting up the other remote checks yesterday.

PR #23, opened 2017-05-28
Replace webmozart/console with mnapoli/silly

padraic:
- [x] Switch to mnapoli/silly
- [x] Set PHP 7.0 as minimum 
- [x] Remove any stragglers in namespaces
- [x] Refactor command handler
- [x] Check for output decoration (e.g. errors)


padraic:
How tied are we to PHP 5.6 support? Looking at the Travis failure, it's a bug somewhere in Silly 1.5 fixed in 1.6 (which requires PHP 7).

theofidry:
Let's go for PHP 7.0 👍

.travis.yml
```diff
@@ -12,16 +12,14 @@ cache:
...
-    - php: 7.1
+    - php: '7.0'
+    - php: '7.1'
+    - php: '7.1'
       env: COMPOSER_FLAGS='--prefer-lowest --prefer-stable'
     - php: hhvm
     - php: nightly
   allow_failures:
-    - php: 5.5
+    - php: '5.5'
```
theofidry: can actually be removed

.travis.yml
```diff
@@ -12,16 +12,14 @@ cache:
...
-    - php: 5.5
-    - php: 5.6
-    - php: 7.0
-    - php: 7.1
-    - php: 7.1
+    - php: '7.0'
+    - php: '7.1'
+    - php: '7.1'
       env: COMPOSER_FLAGS='--prefer-lowest --prefer-stable'
     - php: hhvm
```
theofidry: let's drop HHVM?

composer.json
```diff
@@ -9,17 +9,18 @@
         }
     ],
     "require": {
-        "php": ">=5.5",
-        "webmozart/console": "^1.0-beta5",
+        "php": ">=7.0",
```
theofidry: `~7.0.0 || ~7.1.0`? PHP is not semver and we don't have a guarantee it will work in PHP 7.2 or PHP 8. Although maybe `^7.0` is enough...

composer.json
```diff
@@ -9,17 +9,18 @@
...
         "webmozart/path-util": "^2.3",
         "symfony/filesystem": "^3.2",
         "symfony/process": "^3.2",
         "symfony/finder": "^3.2",
         "padraic/phar-updater": "^1.0.3",
-        "nikic/php-parser": "^3.0"
+        "nikic/php-parser": "^3.0",
+        "mnapoli/silly": "^1.6",
+        "symfony/console": "^3.2"
```
theofidry: do we need to include it? I believe it is already required by `mnapoli/silly`

src/Console/Application.php
```diff
@@ -0,0 +1,36 @@
...
+namespace Webmozart\PhpScoper\Console;
+
+use Silly\Application as SillyApplication;
+
+class Application extends SillyApplication
+{
+    /**
+     * The version of the PHP-Scoper.
+     */
+    const VERSION = '@package_version@';
```
theofidry: Maybe we could play with https://github.com/Ocramius/example-symfony-cli-usage/pull/4 instead which avoids keeping a hard coded constant in sync

src/Handler/AddPrefixCommandHandler.php
```diff
@@ -58,43 +58,65 @@ public function __construct()
      *
      * @return int Returns 0 on success and a positive integer on error.
      */
-    public function handle(Args $args, IO $io)
+    public function handle($prefix, array $paths, BasicFormatter $formatter)
```
theofidry: - phpdoc is outdated
- what's prefix type?
- I would add a phpdoc block for `$paths` as well which I guess is a `string[]`

src/Handler/AddPrefixCommandHandler.php
```diff
@@ -58,43 +58,65 @@ public function __construct()
...
         return 0;
     }
 
-    private function scopeFile($path, $prefix, IO $io)
+    private function scopeFiles($prefix, BasicFormatter $formatter)
+    {
+        $count = count($this->finder);
+        $formatter->outputFileCount($count);
+
+        if ($count > 0) {
```
theofidry: I don't think the count is necessary, the foreach will simply not loop if the count is `0`

theofidry:
@padraic messed up, requires changes instead of approving

padraic:
@theofidry Too late, totally merging it now! ;)

I'll make those changes later.

composer.json
padraic: There was a specific problem with `symfony/console` at 3.0, pegging it to 3.2 resolves that on `--prefer-lowest` builds.

src/Console/Application.php
padraic: I've left it as-is for the time being. It's reasonably complex to figure out given there's a couple of methods php-scoper might end up being installed. An exercise for another day.

.travis.yml
```diff
@@ -12,23 +12,15 @@ cache:
 
 matrix:
   include:
-    - php: 5.5
-    - php: 5.6
-    - php: 7.0
-    - php: 7.1
-    - php: 7.1
-      env: COMPOSER_FLAGS='--prefer-lowest --prefer-stable'
```
theofidry: `--prefer-stable` is unneeded as AFAIK it's the default

.travis.yml
```diff
@@ -12,23 +12,15 @@ cache:
...
   fast_finish: true
 
-install: composer update $COMPOSER_FLAGS --prefer-source -n
+install: composer update --no-interaction --no-suggest --prefer-dist $COMPOSER_FLAGS
 
-script: vendor/bin/phpunit --verbose --coverage-clover=coverage.clover
-
-after_script:
-  - sh -c 'if [ "$TRAVIS_PHP_VERSION" != "hhvm" ]; then wget https://scrutinizer-ci.com/ocular.phar && php ocular.phar code-coverage:upload --format=php-clover coverage.clover; fi;'
+script: vendor/bin/phpunit --verbose --coverage-text
```
theofidry: I would add a `phpenv config-rm xdebug.ini || true` in `before_script` or during `install` and then do `phpdbg -qrr vendor/bin/phpunit --verbose --coverage-text`

composer.json
```diff
@@ -9,17 +9,18 @@
         }
     ],
     "require": {
-        "php": ">=5.5",
-        "webmozart/console": "^1.0-beta5",
+        "php": "^7.0 || ^7.1",
```
theofidry: either `~7.0.0 || ~7.1.0` which is equivalent to `7.{0,1}.*` or `^7.0` which is equivalent to `7.*`

src/Console/Application.php
theofidry: Fair enough :)

padraic:
@theofidry Put those final bits through!

theofidry:
👌

PR #26, opened 2017-06-01
Tests for PHP 7.0 NS syntax

padraic:
Adding PHP 7 syntax for testing. Fails with `MyNs\{Foo,Bar}` style use statements pending a fix.


padraic:
@theofidry Changes should fix up the PHP 7 syntax issue.

theofidry:
Closed in favour of #31

PR #28, opened 2017-06-02
[WIP] Add method of ignoring classes, constants and functions from base PHP and extensions

padraic:
On a parallel to the PHP 7 fix using a custom attribute set on nodes to ignore PHP/SPL/Extension classes. Same is also needed for:

- [ ] Apply to FQN constants, e.g. `\PHP_EOL` vs `PHP_EOL`.
- [ ] Apply to FQN functions, should be even rarer then constants but possible, e.g. `\str_replace()`.

### DIFF
--- a/src/NodeVisitor/IgnoreNamespaceScoperNodeVisitor.php
+++ b/src/NodeVisitor/IgnoreNamespaceScoperNodeVisitor.php
@@ -13,6 +13,7 @@
 namespace Webmozart\PhpScoper\NodeVisitor;
 
 use PhpParser\Node;
+use PhpParser\Node\Name\FullyQualified;
 use PhpParser\Node\Stmt\UseUse;
 use PhpParser\NodeVisitorAbstract;
 
@@ -20,7 +21,7 @@ final class IgnoreNamespaceScoperNodeVisitor extends NodeVisitorAbstract
 {
 
     /**
-     * @var array
+     * @var array Class names to ignore when scoping
      */
     private $reserved;
 
@@ -35,6 +36,9 @@ final class IgnoreNamespaceScoperNodeVisitor extends NodeVisitorAbstract
     public function enterNode(Node $node)
     {
         if ($node instanceof UseUse && in_array((string) $node->name, $this->reserved)) {
+            $node->setAttribute('phpscoper_ignore', true);
+        }
+        if ($node instanceof FullyQualified && in_array((string) $node, $this->reserved)) {
             $node->setAttribute('phpscoper_ignore', true);
         }
     }
--- a/src/NodeVisitor/IgnoreNamespaceScoperNodeVisitor.php
+++ b/src/NodeVisitor/IgnoreNamespaceScoperNodeVisitor.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 /*
  * This file is part of the humbug/php-scoper package.
  *
@@ -10,7 +12,8 @@
  * file that was distributed with this source code.
  */
 
-namespace Webmozart\PhpScoper\NodeVisitor;
+
+namespace Humbug\PhpScoper\NodeVisitor;
 
 use PhpParser\Node;
 use PhpParser\Node\Name\FullyQualified;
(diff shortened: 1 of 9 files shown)


padraic:
Of course, actually compiling a list would be needed. We have PHP (easy), SPL (long but easy), and installed extensions (good luck since we can't predict what is installed). Has anyone out there actually solved this already?

src/NodeVisitor/IgnoreNamespaceScoperNodeVisitor.php
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+namespace Webmozart\PhpScoper\NodeVisitor;
+
+use PhpParser\Node;
+use PhpParser\Node\Name;
+use PhpParser\Node\Stmt\UseUse;
+use PhpParser\NodeVisitorAbstract;
+
+class IgnoreNamespaceScoperNodeVisitor extends NodeVisitorAbstract
```
theofidry: `final`

src/NodeVisitor/IgnoreNamespaceScoperNodeVisitor.php
```diff
@@ -0,0 +1,33 @@
...
+{
+
+    private static $reserved = [
```
theofidry: `private const`? or can be a `const` with the phpdoc `@private`

src/NodeVisitor/UseNamespaceScoperNodeVisitor.php
```diff
@@ -22,6 +22,9 @@ public function __construct($prefix)
     public function enterNode(Node $node)
     {
         if ($node instanceof UseUse) {
+            if ($node->hasAttribute('phpscoper_ignore')) {
```
theofidry: AFAICS it's not specific to the use statement, consider the following:

```php
<?php

namespace A;

function (\Closure $x) {
}
```

or:

```php
<?php

// global namespace

function (Closure $x) {
}
```

theofidry:
To expand on what I mean. If you use `get_declared_classes` at the right time, you can get the list of PHP classes coming from the PHP core + its extensions, so we could take advantage on that by maybe running it as the very first thing in the script process.

padraic:
I'll expand on it a bit tomorrow. Even if this gets us 80% of the way, it's better than 0% and documentation may help explain the difference (extension variation).

src/NodeVisitor/UseNamespaceScoperNodeVisitor.php
padraic: I've left typehinting separate for now. I'll add issues if I don't swing past these and the Slack mentioned items tomorrow evening.

padraic:
Updated description to address constants and functions which have the same problem (via FullyQualified visitor prefixing).

theofidry:
Should we touch the global namespace? I.e. if we have a code:

```php
<?php

namespace {
  class X {
  }
}

namespace Foo {
  class X {
  }
}
```

what should happen? I think the following would be acceptable:

```php
<?php

namespace {
  class X {
  }
}

namespace Humbug\Foo {
  class X {
  }
}
```

but to leave the global namespace alone. Also we will need to account for conflicts at some point:

```php
<?php

namespace Foo {
  class X {
  }
}

namespace Humbug\Foo {
  class X {
  }
}

// Apply prefix with `Humbug`

<?php

namespace Humbug\Foo {
  class X {
  }
}

namespace Humbug\Foo {
  class X {
  }
}
```

which will likely crash

padraic:
@theofidry I'd leave the global namespace, but needs tests to check that through to downstream references (which are probably prefixed at present). Globally namespaced stuff would have just one part, vs 2+ for anything namespaced differently. The solution I started from, building lists upfront such as from `get_declared_classes`, might still be needed as a supplement to that (e.g. extensions using namespaces or if this were to pass: https://wiki.php.net/rfc/namespaces-in-core).

Conflicts: Detection would be half the battle... Basically collect namespace(s), append class refs (constants? functions?) to a list as we scope each file, and monitor for any repeat declaration of these as scoping progresses? Beyond that, one obvious documentation point might be to recommend using a unique prefix (not something shared by code to be scoped such as `Humbug` in your example). I'm not sure if selectively not prefixing things should be allowed - other than the global namespace.

theofidry:
Let's leave the global namespace alone then and not account for conflicts. Maybe we can add a check later to try to circumvent that, but it's not a high priority.

PR #29, opened 2017-06-03
Refactor code + tests

theofidry:
- Given the very little usage we had of `mnapoli/silly` and `webmozart/path-util`, I removed those packages
- I updated the namespace from `Webmozart\PhpScoper` to `Humbug\PhpScoper`
- Added a Make command to run the tests with coverage
- Implemented the check version using `ocramius/package-versions` which is trivial but the output is not satisfying. Waiting on https://github.com/Ocramius/PackageVersions/issues/43 and https://github.com/Ocramius/PackageVersions/issues/44 before tacking action there
- Add scalar typehints as we are now using PHP7

What is missing:

- Apply CS fixes. We need to configure StyleCI to add the missing declare strict statements and license headers with the following license:

```
This file is part of the humbug/php-scoper package.

Copyright (c) 2017 Théo FIDRY <theo.fidry@gmail.com>,
                   Pádraic Brady <padraic.brady@gmail.com>

For the full copyright and license information, please view the LICENSE
file that was distributed with this source code.
```

- The tests are failing: maybe I got things mixed with the mirroring so this need to be fixed

There is much more to come but I think this need to be finished first.


bin/php-scoper
```diff
@@ -10,17 +10,20 @@
...
 if (file_exists($autoload = __DIR__.'/../../../autoload.php')) {
     require_once $autoload;
 } else {
     require_once __DIR__.'/../vendor/autoload.php';
 }
 
-$app = new Application();
-$conf = new ApplicationConfig();
-$conf->configure($app);
+$app = new Application('php-scoper', Versions::getVersion('humbug/php-scoper'));
```
padraic: Add test somewhere for output? Specifically, down the line for phar builds, this should be returning a SemVer compatible string, and the current Ocramius package does not (e.g. uses `@` to append metadata instead of `+`). When building Phars, Box has a similarly incompatible approach by not appending `-dev` for builds which are a tagged version + additional commits.

src/Console/Command/AddPrefixCommand.php
```diff
@@ -0,0 +1,123 @@
...
+
+
+    /**
+     * @inheritdoc
+     */
+    protected function configure()
+    {
+        $this
+            ->setName('add-prefix')
+            ->setDescription('Goes through all the PHP files found in the given paths to apply the given prefix.')
```
padraic: '...to namespaces & FQNs.'?

src/Handler/HandleAddPrefix.php
```diff
@@ -0,0 +1,106 @@
...
+    {
+        $count = count($files);
+        $formatter->outputFileCount($count);
+
+        foreach ($files as $file) {
+            $this->scopeFile($file->getPathName(), $prefix, $formatter);
+        }
+    }
+
+    private function scopeFile(string $path, string $prefix, ConsoleLogger $formatter)
```
padraic: Should we rename $formatter var?

src/Scoper.php
```diff
@@ -33,12 +30,14 @@ public function __construct(Parser $parser)
...
+     * @param string $content Content of the file to scope
+     * @param string $prefix Prefix to apply to the file
+     *
+     * @throws ParsingException
      *
-     * @return string
+     * @return string Content of the file with the prefix applied
      */
-    public function addNamespacePrefix($content, $prefix)
+    public function addNamespacePrefix(string $content, string $prefix): string
```
padraic: I'm assuming not declaring strict at top of file is intended (here and a few other files)?

tests/Console/Command/AddPrefixCommandTest.php
```diff
@@ -0,0 +1,254 @@
...
+            return;
+        }
+
+        $this->cwd = getcwd();
+
+        $application = new Application('php-scoper-test');
+        $application->addCommands([
+            new AddPrefixCommand(),
+        ]);
+        $application->setAutoExit(false);
```
padraic: `$application->setCatchExceptions(false);` to not swallow exceptions where these would normally bubble up to output in use.

tests/Console/Command/AddPrefixCommandTest.php
```diff
@@ -0,0 +1,254 @@
...
+
+        $expected = <<<EOF
+Scoping $basePath/MyClass.php. . . Success
+Scoping $basePath/MySecondClass.php. . . Success
+Scoping $basePath/MyThirdClass.php. . . Success
+
+EOF;
+        $expected = str_replace('/', DIRECTORY_SEPARATOR, $expected);
+
+        Assert::assertEmpty(
```
padraic: Why import `Assert` if we're already extending `TestCase`? It seems an oddly roundabout way of using assertions unless it's for the documentation effect?

tests/Console/Command/AddPrefixCommandTest.php
```diff
@@ -0,0 +1,254 @@
...
+Scoping $this->tempDir/dir/MyIncorrectClass.php. . . Fail
+
+EOF;
+        $expected = str_replace('/', DIRECTORY_SEPARATOR, $expected);
+
+        Assert::assertSame(0, $this->appTester->getStatusCode());
+        Assert::assertStringEndsWith($expected, $this->appTester->getDisplay(true));
+    }
+
+    public function test_cannot_scope_nonexistent_files()
```
padraic: There should be an exception floating through here somewhere? It's being suppressed through the ApplicationTester config which isn't clear from this test.

bin/php-scoper
theofidry: The output of `Versions::getVersion()` is `<version>@<hash>`, maybe we could stick to `<version>` if the version is a tagged one and `<hash>` otherwise.

For the test I think we should have one test somewhere with the complete output. That said, I think we should approach the tests differently. Indeed right now we are testing the output each time which has two drawbacks:

- It is tedious, especially when some dynamic parameters comes e.g. when you start to display "Too X time and Y MB"
- The output is the thing the most likely to change like any UI making the tests more painful as any change in the UI impacts a lot of tests
- It is not really what we want to test: we are interested in the output in a few tests, because that's what we want to test, otherwise we only want to test the scoping

src/Handler/HandleAddPrefix.php
theofidry: But is it really a formatter? It feels more like a (non PSR-3 true) logger to me. The output could very well be something different than stdout but a file or something, and we use it only to record things that happened

src/Scoper.php
theofidry: No it was just missing and I forgot to add it, need to be added everywhere

tests/Console/Command/AddPrefixCommandTest.php
theofidry: Because it's a static method so it's supposed to be called that way. Also IIRC at some point it was discussed to no longer have `$this->assert*` in PhpUnit (there was a deprecation). I guess it was reverted.

That said I don't feel strongly about it so I can revert that if looks too weird.

tests/Console/Command/AddPrefixCommandTest.php
theofidry: There should be a try/catch in the code, it's easier to test the status code and previous exception

theofidry:
@padraic do you see why the `fixtures/original` get modified instead of the mirrored files?

theofidry:
@padraic found a nicer way to separate the tests which should make it easier for maintenance. Will push something during the week. Due to how the tests are, I think any change done in https://github.com/humbug/php-scoper/pull/28 will be easy to integrate.

tests/Console/Command/AddPrefixCommandTest.php
padraic: I think the reason PHPUnit backed off deprecating was simple convention - I haven't seen Assert used before directly. Nevertheless it's a negligible issue.

padraic:
@theofidry No worries, take your time. A lot of the other stuff I've committed is a few minutes to transfer so don't worry about that being unsynced.

theofidry:
@padraic an overview of the changes:

- The console command is kept as possible, it validates the input and pass it down to the handler
- The console has two tests: a unit test to really test what it does and an integration test to have 1 e2e test
- The handler does goes through each file, scope & dump each of them. The file validation, i.e. if the file is readable or exists is done there
- The handle now has tests

The biggest difference is I believe in the tests where I refactored the code in such a way it ease the testing for each of the class above, while keeping a very simple e2e test.

There is a lot more to do and that I want to do, but I think this PR is big enough

padraic:
Appears we can close #19 with this also.

PR #34, opened 2017-06-07
Add method of ignoring classes, constants and functions from global namespace

padraic:
Reset of original PR. On a parallel to the PHP 7 fix using a custom attribute set on nodes to ignore PHP/SPL/Extension classes. Same is also needed for:

- [x] Apply to FQN constants, e.g. \PHP_EOL vs PHP_EOL.
- [x] Apply to FQN functions, should be even rarer then constants but possible, e.g. \str_replace().
- [x] Test to non-FQNs to rule out impacts there.
- [x] Test for return type declarations for consistency.

Original used name collection, this instead switches to a simple convention of skipping single part FQNs or non-FQNs. e.g. `\Something` or `Something` vs `\MyNS\Something` or `MyNS\Something`. For now that suggests a need to allow configuration of exceptions for userland code which is part of the global namespace (presumably legacy code/packages).


.styleci.yml
```diff
@@ -10,4 +10,4 @@ disabled:
 
 finder:
   exclude:
-    - 'fixtures'
+    - 'reserved_classes.php'
```
padraic: Exclude not working at present, will need to check or just integrate into the test as a string.

tests/Fixtures/reserved_classes.php
```diff
@@ -0,0 +1,15 @@
+<?php
```
theofidry: I think you could move that under `ScoperTest:: provideValidFiles()`

src/NodeVisitor/FullyQualifiedNamespaceUseScoperNodeVisitor.php
```diff
@@ -37,6 +37,11 @@ public function __construct($prefix)
     public function enterNode(Node $node)
     {
         if ($node instanceof FullyQualified) {
```
theofidry:
```php
if (
    $node instanceof FullyQualified
    (false === $node->hasAttribute('phpscoper_ignore') || false === $node->getAttribute('phpscoper_ignore'))
) {
    return new Name(Name::concat($this->prefix, (string) $node));
}
```

src/NodeVisitor/UseNamespaceScoperNodeVisitor.php
```diff
@@ -38,9 +38,15 @@ public function __construct($prefix)
     public function enterNode(Node $node)
     {
         if ($node instanceof UseUse
-            && $node->hasAttribute('parent') && false === ($node->getAttribute('parent') instanceof  GroupUse)
+            && $node->hasAttribute('parent')
+            && false === ($node->getAttribute('parent') instanceof GroupUse)
             && $this->prefix !== $node->name->getFirst()
         ) {
+            if ($node->hasAttribute('phpscoper_ignore')
```
theofidry: like above I think this can be moved to a single if statement

theofidry:
Some small CS stuff nitpicking, all good otherwise

PR #35, opened 2017-06-09
Pretty output

theofidry:


theofidry:
@padraic not sure to understand why AppVeyor is failing...

padraic:
@theofidry The Appveyor failure is a missing function import somewhere in PS. It's either a problem with the PHP package (which should import it) or with Choco. I'd make a tiny edit to `appveyor.yml` (invalidates the cache, e.g. reorder the `init` section lines which is harmless) and that should trigger a clean install.

tests/Console/Command/AddPrefixCommandIntegrationTest.php
```diff
@@ -99,6 +99,120 @@ public function test_scope_the_given_paths()
...
+ / ____/ __  / ____/   ___/ / /__/ /_/ / /_/ /  __/ /    
+/_/   /_/ /_/_/       /____/\___/\____/ .___/\___/_/     
+                                     /_/
+
+PHP Scoper version 12ccf1ac8c7ae8eaf502bd30f95630a112dc713f
+
+ * [OK] /path/to/file.php
+
+
+ [OK] Successfully prefixed 1 files.
```
padraic: Test for failure output? Moreso to complete a description that we expect the app to return an `FA` line, print memory stats, and simply print an exception before quitting - though really, the exception being printed is the main objective for user feedback here.

theofidry: I though so as well but as I'm thinking of bailing on error in #36 so we would get more useful feedback there

theofidry:
@padraic should I wait for AppVeyor or can I merge it?

padraic:
@theofidry You can probably just merge - if it hasn't already cleared the queue. No parallel race issues I know of.

PR #36, opened 2017-06-09
Transactional scoping and artefact generation

theofidry:


theofidry:
@padraic I'm a bit blocked on one thing. So I'm trying to dump the scoped files in an output directory. I was thinking of:

```
input-dir/
  file1.php
  file2.php

=> scope files in the output dir `output-dir`

output-dir/
  file1.php
  file2.php
```

On paper it looks good, but in practice we have multiple paths in input. So should we do something like that?

```
input-dir/
  file1.php
  file2.php

/another-path/another-input/dir/
  fileA.php
  fileB.php

=> scope files in the output dir `output-dir`

output-dir/
  path/to/input-dir
    file1.php
    file2.php
  another-path/another-input/dir/
    fileA.php
    fileB.php
```

padraic:
Option 2 looks fine here, if we can establish that both path references have a common root we can use to set the relative path within `output` (e.g. a `vendor` and `vendor-bin` at root of a package)? It depends on what the use case actually is. I'm assuming a simple case of isolating a single package's deps where the package root is obvious.

```
/path/to/input-dir/
  file1.php
  file2.php
/path/to2/another-input
  fileA.php
  fileB.php

=> scope files in the output dir `output-dir`

output-dir/
  to/input-dir
    file1.php
    file2.php
  to2/another-input
    fileA.php
    fileB.php
```

So in the above `/path` is the common root, so the relative from that point sets output. Then, of course, document that as the behaviour before someone uses two wildly different paths from unrelated packages/locations, and expects something else.

On another note, we should work through an example of building a phar and check through the actual workflow, problems, advantages/disadvantages. I wouldn't encase php-scoper's default behaviour in concrete until then.

theofidry:
After #37 I indeed think °2 is the solution as you're likely to give multiple paths...

Issue #5, opened 2015-10-20
Add --fix-composer option


webmozart:
The option `--fix-composer` should be added which adds the namespace prefix to the "autoload" section of all composer.json files found in the searched paths.

theofidry:
We now scope the `composer.json` files to fix the autoloading of the source code and the `installed.json` of composer for the autoloading. No option needed.

PR #38, opened 2017-06-11
[DO NOT MERGE] Prototype of the first MVP

theofidry:
I played a bit with this library in the train to see what is really missing and attempt to make some progress. Whilst this PR should not be merged directly, some of its changes should be cherry-picked and merge via smaller and more scoped PRs.

<hr>

What the main changes are about:

- Change the build (for generating the PHAR) process of php-scoper to
apply itself in order to isolate its deps
- Tweak the command to make it more usable and fix some stuffs
- Scope all files even non-PHP one (or rather dump them as well, they
don't need to be isolated)
- Show file path when a parse error occurrs
- Fix progress bar

@padraic I'm quite satisfied overall and I don't think we are missing much for 1.0.0. The tweaks done here are the mostly it, the only blocking thing I see for now is:

- Fixing the namespacing in strings, necessary at least for the Composer autoloading otherwise the classmap used is wrong
- Allowing the whitelisting of some namespaces/classes to make it usable with applications like PhpUnit

We will need a proper test for the second case at some point. For the first one, a good test is to actually run the build process of the PHAR for php-scoper itself where in this PR I apply the scoping first, and build the PHAR from the scoped files.


theofidry:
Note that an alternative to fixing the classmap directly is to fix the `composer.json` like suggested in #5 which can actually may turn out to be much easier.

padraic:
I'd certainly go with fixing composer.json and see if that suffices.

theofidry:
Wouldn't that require to dump the autoloader again though?

theofidry:
Closed in favour of #40, #41, #42

PR #40, opened 2017-06-12
Enhance usability of the command

theofidry:
- Make prefix argument an optional option which defaults to a random prefix
- Add prefix to all the files found in the current working directory when no path given
- Throw an error when the output directory already exists and is not writeable
- Add a force option for non-interactive mode to be able to add prefixes although the output directory already exists and is not empty (without the option the process is interrupted)
- Make the errors a bit more readable by adding a color to the file paths
- Mock the filesystem in the command unit tests to control the calls being made and ensure no `remove()` is called by mistake and change the output dir to something different than the project dir... cost me a few project deletion


theofidry:
For some reasons as you can see on https://scrutinizer-ci.com/g/humbug/php-scoper/inspections/fff4a46a-80ce-400a-95ea-087c6c821763, sometimes PhpUnit hangs out for no reason...

"Funny" fact, it does the same for me locally but works fine when I really on PhpStorm PhpUnit (which should use the same binary from what I see in my config...)

padraic:
Odd, I don't see any changes that stand out to create the problem. You can't argue with phpunit though! PHPUnit can be configured to set a time limit on individual tests that might let you weed out the offender.

theofidry:
The biggest problem is that I need to do step by step debugging or have a profile to see what's going one, but everything works fine when running the tests from PhpStorm and profiling fails because of the timeout. Do you have the same issue locally?

padraic:
I can't check locally at the moment unfortunately, will be out for most of the afternoon too.

theofidry:
@padraic found the offender, it was because even though I was in the context of PhpUnit, my shell was interactive so it was waiting for an input for the questions asking confirmation to delete a directory

src/Console/Command/AddPrefixCommand.php
```diff
@@ -59,22 +61,29 @@ protected function configure()
...
                 'o',
                 InputOption::VALUE_REQUIRED,
                 'The output directory in which the prefixed code will be dumped.',
                 'build'
             )
+            ->addOption(
+                self::FORCE_OPT,
+                'f',
+                InputOption::VALUE_NONE,
+                'Deletes any existing content in the output directory without any warning'
```
padraic: Insert `rm -rf /` joke ;).

Do we test that the output directory is removed somewhere (esp. for Windows where Symfony Filesystem appears to have issues with this)?

theofidry: Err not really. I think integration tests would be better here given how FS can be troublesome, but there's some cases where mocking the FS would be more appropriate (e.g. when you change the perm to test the `is_writeable()`. I kinda wasted quite some time with that FS thing which is why I dropped it and checked manually for now

theofidry:
We'll need to fix AppVeyor at some point as well although less urgent IMO

src/Console/Command/AddPrefixCommand.php
padraic: Fair enough. Merge away!

Issue #39, opened 2017-06-12
Check namespace aliases for single level renames?


padraic:
While pointless single level namespaces will raise a warning in use, there is an exception for setting an alias. As equally pointless as that may be...
```
<?php
use Symfony; // raise warning
use Symfony as Laravel; // no warning
new Laravel\Component\Finder\Finder; // Symfony Finder instance
```
With the most recent change in scoping, any `use XXX as YYY` would escape scoping for now.

theofidry:
@padraic I would generate a unique alias then, but 👍

padraic:
Closed via #50

PR #50, opened 2017-06-14
Check if single level use statement is an aliased namespace, not simply a global namespace class

padraic:
Resolution to #39


src/Scoper.php
```diff
@@ -50,6 +51,7 @@ public function scope(string $content, string $prefix): string
     {
         $traverser = new NodeTraverser();
         $traverser->addVisitor(new ParentNodeVisitor());
+        $traverser->addVisitor(new SingleLevelUseAliasVisitor($prefix));
```
theofidry: shouldn't it come after the ignore namespace?

padraic: I don't think it matters since the enterNode() needs the check a UseUse Name child node - the order is bound to parent first, child second no matter the order of the visitors.

PR #51, opened 2017-06-14
Add a basic SelfUpdateCommand using phar-updater v1

padraic:
Implements #45 

Assumes using Github releases to host a signed phar.

### DIFF
--- a/src/Console/Command/SelfUpdateCommand.php
+++ b/src/Console/Command/SelfUpdateCommand.php
@@ -14,11 +14,13 @@ declare(strict_types=1);
 
 namespace Humbug\PhpScoper\Console\Command;
 
+use Humbug\PhpScoper\Logger\UpdateConsoleLogger;
 use Humbug\SelfUpdate\Updater;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
 
 final class SelfUpdateCommand extends Command
 {
@@ -34,6 +36,11 @@ final class SelfUpdateCommand extends Command
     const CHECK_OPT = 'check';
 
     /**
+     * @var Updater
+     */
+    private $updater;
+
+    /**
      * @var OutputInterface
      */
     private $output;
@@ -43,16 +50,38 @@ final class SelfUpdateCommand extends Command
      */
     private $version;
 
+    /**
+     * @var UpdateConsoleLogger
+     */
+    private $logger;
+
+    /**
+     * @param Updater $updater
+     */
+    public function __construct(Updater $updater)
+    {
+        parent::__construct();
+
+        $this->version = $this->getApplication()->getVersion();
+        $this->updater = $updater;
+    }
+
+    /**
+     * @inheritdoc
+     */
     protected function configure()
     {
         $this
             ->setName('self-update')
-            ->setDescription('Update php-scoper.phar to most recent stable build.')
+            ->setDescription(sprintf(
+                    'Update %s to most recent stable build.',
+                    $this->getLocalPharName()
+            ))
             ->addOption(
                 self::ROLLBACK_OPT,
                 'r',
                 InputOption::VALUE_NONE,
-                'Rollback to previous version of php-scoper if available on filesystem.'
+                'Rollback to previous version of PHP-Scoper if available on filesystem.'
             )
             ->addOption(
                 self::CHECK_OPT,
@@ -64,13 +93,17 @@ final class SelfUpdateCommand extends Command
     }
 
     /**
-     * @param InputInterface  $input
-     * @param OutputInterface $output
+     * @inheritdoc
      */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
+        $io = new SymfonyStyle($input, $output);
+        $this->logger = new UpdateConsoleLogger(
+            $this->getApplication(),
+            $io
+        );
+
         $this->output = $output;
-        $this->version = $this->getApplication()->getVersion();
 
         if ($input->getOption('rollback')) {
             $this->rollback();
@@ -97,55 +130,40 @@ final class SelfUpdateCommand extends Command
      */
     private function getStableUpdater(): Updater
     {
-        $updater = new Updater();
-        $updater->setStrategy(Updater::STRATEGY_GITHUB);
+        $this->updater->setStrategy(Updater::STRATEGY_GITHUB);
         return $this->getGithubReleasesUpdater($updater);
     }
 
     private function update(Updater $updater)
     {
-        $this->output->writeln('Updating...'.PHP_EOL);
+        $this->logger->startUpdating();
         try {
-            $result = $updater->update();
+            $result = $this->updater->update();
 
-            $newVersion = $updater->getNewVersion();
-            $oldVersion = $updater->getOldVersion();
+            $newVersion = $this->updater->getNewVersion();
+            $oldVersion = $this->updater->getOldVersion();
         
             if ($result) {
-                $this->output->writeln('php-scoper has been updated.');
-                $this->output->writeln(sprintf(
-                    'Current version is: %s.',
-                    $newVersion
-                ));
-                $this->output->writeln(sprintf(
-                    'Previous version was: %s.',
-                    $oldVersion
-                ));
+                $this->logger->updateSuccess($newVersion, $oldVersion);
             } else {
-                $this->output->writeln('php-scoper is currently up to date.');
-                $this->output->writeln(sprintf(
-                    'Current version is: %s.',
-                    $oldVersion
-                ));
+                $this->logger->updateNotNeeded($oldVersion);
             }
         } catch (\Exception $e) {
-            $this->output->writeln(sprintf('Error: %s', $e->getMessage()));
+            $this->logger->error($e);
         }
-        $this->output->write(PHP_EOL);
     }
 
     private function rollback()
     {
-        $updater = new Updater();
         try {
-            $result = $updater->rollback();
+            $result = $this->updater->rollback();
             if ($result) {
-                $this->output->writeln('php-scoper has been rolled back to prior version.');
+                $this->logger->rollbackSuccess();
             } else {
-                $this->output->writeln('Rollback failed for reasons unknown.');
+                $this->logger->rollbackFail();
             }
         } catch (\Exception $e) {
-            $this->output->writeln(sprintf('Error: %s', $e->getMessage()));
+            $this->logger->error($e);
         }
     }
 
@@ -157,10 +175,7 @@ final class SelfUpdateCommand extends Command
 
     private function printCurrentLocalVersion()
     {
-        $this->output->writeln(sprintf(
-            'Your current local version is: %s',
-            $this->version
-        ));
+        $this->logger->printLocalVersion($this->version);
     }
 
     private function printCurrentStableVersion()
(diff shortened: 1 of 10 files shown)


src/Console/Command/SelfUpdateCommand.php
```diff
@@ -0,0 +1,205 @@
...
+    /**
+     * @var string
+     */
+    private $version;
+
+    protected function configure()
+    {
+        $this
+            ->setName('self-update')
+            ->setDescription('Update php-scoper.phar to most recent stable build.')
```
theofidry: can the name be changed? I mean the user may have renamed `php-scoper.phar` to something else e.g. `php-scoper` or `scoper`

src/Console/Command/SelfUpdateCommand.php
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    /**
+     * @param InputInterface  $input
+     * @param OutputInterface $output
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $this->output = $output;
+        $this->version = $this->getApplication()->getVersion();
```
theofidry: can this be done in the constructor?

src/Console/Command/SelfUpdateCommand.php
```diff
@@ -0,0 +1,205 @@
...
+    {
+        $this->update($this->getStableUpdater());
+    }
+
+    /**
+     * @return Updater
+     */
+    private function getStableUpdater(): Updater
+    {
+        $updater = new Updater();
```
theofidry: I didn't check the updater API but is there any reason why the strategy cannot be se in the updater constructor?

src/Console/Command/SelfUpdateCommand.php
```diff
@@ -0,0 +1,205 @@
...
+    {
+        $this->output->writeln('Updating...'.PHP_EOL);
+        try {
+            $result = $updater->update();
+
+            $newVersion = $updater->getNewVersion();
+            $oldVersion = $updater->getOldVersion();
+        
+            if ($result) {
+                $this->output->writeln('php-scoper has been updated.');
```
theofidry: PHP-Scoper

src/Console/Command/SelfUpdateCommand.php
```diff
@@ -0,0 +1,205 @@
...
+                $this->output->writeln(sprintf(
+                    'Current version is: %s.',
```
theofidry: `<comment>%s</comment>`

src/Console/Command/SelfUpdateCommand.php
```diff
@@ -0,0 +1,205 @@
...
+                    $newVersion
+                ));
+                $this->output->writeln(sprintf(
+                    'Previous version was: %s.',
```
theofidry: `<comment>%s</comment>`

src/Console/Command/SelfUpdateCommand.php
```diff
@@ -0,0 +1,205 @@
...
+                $this->output->writeln('php-scoper is currently up to date.');
+                $this->output->writeln(sprintf(
+                    'Current version is: %s.',
+                    $oldVersion
+                ));
+            }
+        } catch (\Exception $e) {
+            $this->output->writeln(sprintf('Error: %s', $e->getMessage()));
+        }
+        $this->output->write(PHP_EOL);
```
theofidry: `newLine()` or `writeln('')`

src/Console/Command/SelfUpdateCommand.php
```diff
@@ -0,0 +1,205 @@
...
+                ));
+            } else {
+                $this->output->writeln('php-scoper is currently up to date.');
+                $this->output->writeln(sprintf(
+                    'Current version is: %s.',
+                    $oldVersion
+                ));
+            }
+        } catch (\Exception $e) {
+            $this->output->writeln(sprintf('Error: %s', $e->getMessage()));
```
theofidry: is the rollback done in the updater in case of failure? If so we don't actually need to catch the exception as we'll have more information by throwing letting the Console handling the exception

src/Console/Command/SelfUpdateCommand.php
theofidry: can we create the updater in the constructor or better use DI (for mocking in unit tests)?

padraic:
@theofidry Yeah, still a bit of work to tidy this up. Will be tomorrow though :). The output can probably go to a logger somewhere too.

src/Console/Command/SelfUpdateCommand.php
padraic: The basic update flow is to backup the original phar (leaving original still in place), download the new phar to a temporary directory, validate it, and then copy it to the original phar location (overwrite). On error, the temporary copy would be deleted. The rollback is largely in the event that an update is valid, but broken or incompatible for any reason and you want the old version back.

src/Console/Command/SelfUpdateCommand.php
padraic: `Phar::running()` should return the file path with the local name, I think.

padraic:
@theofidry Will look into testing tomorrow. Just cleaned it up and simplified.

src/Console/Command/SelfUpdateCommand.php
```diff
@@ -0,0 +1,198 @@
...
+        if ($input->getOption('rollback')) {
+            $this->rollback();
+
+            return;
+        }
+
+        if ($input->getOption('check')) {
+            $this->printAvailableUpdates();
+
+            return;
```
theofidry:
```php
return 0;
```

src/Console/Command/SelfUpdateCommand.php
```diff
@@ -0,0 +1,198 @@
...
+        }
+
+        $this->update($this->getStableUpdater());
+    }
+
+    /**
+     * @return Updater
+     */
+    private function getStableUpdater(): Updater
```
theofidry: `createUpdater()`?

theofidry: Also I think the phpdoc is unneeded here

src/Console/Command/SelfUpdateCommand.php
```diff
@@ -0,0 +1,198 @@
...
+
+            $newVersion = $this->updater->getNewVersion();
+            $oldVersion = $this->updater->getOldVersion();
+
+            if ($result) {
+                $this->logger->updateSuccess($newVersion, $oldVersion);
+            } else {
+                $this->logger->updateNotNeeded($oldVersion);
+            }
+        } catch (\Exception $e) {
```
theofidry: What kind of error can we get? If it's a generic one maybe we should use `Throwable` instead?

padraic:
Made a few updates. For reference, the local `box.json` adds the following new fields:

```js
    "git-version": "git-version",
    "algorithm": "OPENSSL",
    "key": "/path/to/your/private.pem",
    "key-pass": true
```

Box version we use is pointing at Symfony/Console 3.x as an option, but the underlying commands are not updated to use the Question helper (still using Dialog). Adjusted our own `vendor-bin/composer.json` to fix.

I may temporarily hijack the 0.1.0 Github release for practical download/rollback testing tomorrow.

The phar will emit a semantic version based off of box's `git-version`, so it will be one of:

- `x.y.z` (typical stable version when phar is build to a tagged commit.
- `x.y.z+15-g0818afa (where `+` marks anything after as metadata, `15` is the commit count since previous tagged commit, and `0818afa` is the current short commit hash prefix with `g` by box.

theofidry:
>Box version we use is pointing at Symfony/Console 3.x as an option, but the underlying commands are not updated to use the Question helper (still using Dialog). Adjusted our own vendor-bin/composer.json to fix.

I'm not following. You mean it allows `symfony/console 3.x` but actually the code is still using `2.x` code which as been removed in `3.x`?

Nevermind, actually saw [that](https://github.com/humbug/php-scoper/pull/51/files#diff-9640169ec654daef753be064011c0e27R10) which confirms it. Seems fine by me really 👍 

I'm must stay I'm still not sure how the phar updater works exactly as I'm only used to the Composer one.

padraic:
@theofidry That's the problem - still using Symfony 2's Dialog helper in the background.

The updater works a lot like Composer's - I doubt there's a huge difference between them at the basic level. The differences are all extras to the basic workflow: phar signature support, the strategies (all essentially varying methods of where/how to check for updates, and where to download any updates).

src/Console/Command/SelfUpdateCommand.php
padraic: Null is converted to 0 status automatically isn't it? No harm being explicit though.

src/Console/Command/SelfUpdateCommand.php
padraic: Any day now I'll wake up and remember PHP 7 was released ;).

padraic:
@theofidry All wrapped up here!

src/Console/Command/SelfUpdateCommand.php
```diff
@@ -0,0 +1,196 @@
...
+            $newVersion = $this->updater->getNewVersion();
+            $oldVersion = $this->updater->getOldVersion();
+
+            if ($result) {
+                $this->logger->updateSuccess($newVersion, $oldVersion);
+            } else {
+                $this->logger->updateNotNeeded($oldVersion);
+            }
+        } catch (\Throwable $e) {
+            $this->logger->error($e);
```
theofidry: we're not failing here?

src/Console/Command/SelfUpdateCommand.php
```diff
@@ -0,0 +1,196 @@
...
+                $this->logger->currentVersionInstalled($stability);
+            }
+        } catch (\Throwable $e) {
+            $this->logger->error($e);
+        }
+    }
+
+    private function getLocalPharName(): string
+    {
+        return basename(\PHAR::running());
```
theofidry: This command doesn't work without a PHAR, but can we ensure it fails gracefully?

src/Logger/UpdateConsoleLogger.php
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    public function startUpdating()
+    {
+        $this->io->writeln('Updating...');
+    }
+
+    public function updateSuccess(string $newVersion, string $oldVersion)
+    {
+        $this->io->writeln('PHP-Scoper has been updated.');
```
theofidry: with `SymfonyStyle` you also have `success()`, `error()` and such for some fancy styling

src/Console/Command/SelfUpdateCommand.php
padraic: Mostly, it's just to add some preceding text to assure users the phar was not replaced. We could let the Throwable just fall through, but I always find that a bit untidy.

theofidry: Not sure about this though as we rule out any way for the user to know why it failed. I think throwing the throwable after logging the message would be fine

theofidry:
Hum Travis is now failing because of the missing pubkey...

padraic:
Hmm, phar-updater checks it up front instead of waiting for a request for a pub key to run checks: https://github.com/humbug/phar-updater/blob/master/src/Updater.php#L112

It should be changed in the updater package for these scenarios (unsigned testing). I'll open an issue there. In the meantime, we can probably suppress that exception. For actual updating in RL, the phar itself won't even run if the pubkey were missing.

src/Console/Command/SelfUpdateCommand.php
padraic: Can you clarify? The SelfUpdateCommand only ever exists in a PHAR, so the above `running()` call should always succeed.

padraic:
@theofidry All comments should be addressed now. Made one change to makefile so we ignore, only during tests, any local `box.json` (e.g. with privkey config) and use the `dist` copy instead.

padraic:
@theofidry Made a couple of changes. The remainder on Scrutinizer we can't fix from outside phar-updater.

PR #52, opened 2017-06-15
Scope Composer autoloader

### LINKED ISSUE #46: Handle Composer autoloading
Follow up of https://github.com/humbug/php-scoper/pull/38#issuecomment-307658139

theofidry:
Closes #46.

Right is at a state of a POC.

What is missing:

- Add tests
- Run `composer dump-autoload` after the scoping (I ran it manually for my tests)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,7 @@ build: bin/php-scoper
 	rm composer.lock
 	composer install --no-dev --prefer-dist --classmap-authoritative
 	php -d zend.enable_gc=0 bin/php-scoper add-prefix --force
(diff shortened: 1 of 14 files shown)


theofidry:
@padraic should we try to run the autoload dump now or we can leave it to the user until we find a good way to do that?

padraic:
@theofidry I vote leave it for the user for now. And perhaps in the future add an opt-in option to command?

theofidry:
Let's do that then, let's get something working well first and then work more on the DX.

PR #56, opened 2017-06-15
Properly scope fully-qualified class name nodes

### LINKED ISSUE #53: Scope files in the global namespace
I had a simple file:

```php
<?php

require_once __DIR__.'/vendor/autoload.php';

echo (new \Foo\Foo())().PHP_EOL;
```

Where `Foo` was an invokable. Surprisingly the scoping did not prefix `\Foo\Foo()` to `PhpScoperPrefix\Foo\Foo()`.

padraic:
Also fix unrelated bug stripping `\` from start of FQN after concatenating prefixes
closes #53

### DIFF
--- a/src/NodeVisitor/FullyQualifiedNamespaceUseScoperNodeVisitor.php
+++ b/src/NodeVisitor/FullyQualifiedNamespaceUseScoperNodeVisitor.php
@@ -40,7 +40,7 @@ final class FullyQualifiedNamespaceUseScoperNodeVisitor extends NodeVisitorAbstr
             && false === ($node->hasAttribute('phpscoper_ignore')
             && true === $node->getAttribute('phpscoper_ignore'))
         ) {
-            return new Name(Name::concat($this->prefix, (string) $node));
+            return new Name('\\' . Name::concat($this->prefix, (string) $node));
         }
 
         return $node;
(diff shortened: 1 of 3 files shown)


src/NodeVisitor/FullyQualifiedNamespaceUseScoperNodeVisitor.php
```diff
@@ -40,7 +40,7 @@ public function enterNode(Node $node)
             && false === ($node->hasAttribute('phpscoper_ignore')
             && true === $node->getAttribute('phpscoper_ignore'))
         ) {
-            return new Name(Name::concat($this->prefix, (string) $node));
+            return new Name('\\'.Name::concat($this->prefix, (string) $node));
```
padraic: Casting Node to string omits `\` at start so the new Node isn't registered as fully qualified.

tests/Scoper/PhpScoperTest.php
```diff
@@ -594,6 +594,28 @@ public function provideValidFiles()
...
         //
+        // FQN usage for a name
+        //
+        // ====================
+
+        yield '[FQN usage for a name] fully qualified class' => [
+            <<<'PHP'
+<?php
+
+new \Foo\Bar();
```
theofidry: How are we able to tell the difference between `\Closure` and `\Foo`?

theofidry:
Nvm: forgot we are not messing with the global namespace yet so if you have `\Foo` then it's not prefixed

padraic:
Not yet, and the single part visitor I wrote is restricted to handling aliases. What I should do...is open an issue to track.

PR #58, opened 2017-06-15
Add Blackfire profiling command

### LINKED ISSUE #48: Performance
Unlikely to have big improvements but I think having one or two simple scenario like done in https://github.com/nelmio/alice/tree/master/profiling is a good thing just to keep track of it to ensure we don't fuck up badly.

theofidry:
Closes #48. FYI here's what we have right now:

https://blackfire.io/profiles/48ea61f2-d103-46a9-9f43-9ba5560c241d/graph

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,6 +48,22 @@ e2e: vendor
 	php build/set004/bin/greet.phar > build/output
 	diff fixtures/set004/expected-output build/output
 
+tb:				  ## Run Blackfire profiling
+tb: vendor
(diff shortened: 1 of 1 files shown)


padraic:
Anything popping out...other than PHP-Parser ;).

theofidry:
For now yeah not much more, although with the test I did we could measure the impact of Xdebug which is about 30%, so might be worth having a trick a la Composer to disable it at some point

PR #59, opened 2017-06-15
Fix Composer autoloading

theofidry:
Another stab at #46.

@padraic I'm a bit blocked really, seems like Composer gets its metadata from somewhere else. I hoped fixing `composer.lock` would be enough but that's not the case as I can still see:

```php
<?php

// autoload_psr4.php @generated by Composer

$vendorDir = dirname(dirname(__FILE__));
$baseDir = dirname($vendorDir);

return array(
    'PhpScoper59431c8baf5d3\\Set005\\' => array($baseDir . '/src'),
    'Assert\\' => array($vendorDir . '/beberlei/assert/lib/Assert'),
);

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ PHPUNIT=vendor/bin/phpunit
 PHPSCOPER=bin/php-scoper.phar
 
 .DEFAULT_GOAL := help
-.PHONY: build test tu tc e2e
+.PHONY: build test tu tc e2e tb
 
 
 help:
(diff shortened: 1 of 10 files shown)


theofidry:
Should work now. When I try to run the e2e tests with `php-scoper.php`, it works fine, but not with `php-scoper.phar` so there's still an issue remaining. It however should be handled in a separate PR.

As the issue only affects the PHAR, I think we can tag this lib once this PR is merged

padraic:
@theofidry Should have checked before approving, but PHP 7.0 Travis build is broken. Set it to rebuild to double check.

theofidry:
@padraic I meant dev version for the tagging yes

The failure is "expected", the e2e tests are using `php-scoper.phar` instead of `php-scoper.php`. So we either revert to the `.php` to make the test green or figure out what the issue is

padraic:
Yep, that's fine for now. Good to merge.

PR #60, opened 2017-06-17
Set executable perm when building phar

padraic:
Small fix to make phar builds executable under Linux.


padraic:
File limit is also borked, but that can be documented somewhere. Unlike building with box.phar, using box as a dependency seems to raise a completely different exception for the same limit issue.

theofidry:
Maybe we are using the wrong version? Also we're using a fork :/

padraic:
It just seems odd, like the local dependency is catching and rethrowing the original exception, but the phar isn't. Not that big of a deal - could just add a warning to the makefile to document the error is expected with doing X.

Issue #63, opened 2017-06-18
Handle custom string manipulations


padraic:
@theofidry I checked briefly into the PHAR problems, and it seems to boil down to PHP Parser referring to specific classes and functions using strings (which we aren't scoping). Two examples:

`PhpParser\Lexer::createTokenMap()`:
```php
if (defined($name = 'PhpParser\\Parser\\Tokens::' . $name)) {
    // Other tokens can be mapped directly
    $tokenMap[$i] = constant($name);
}
```
Straightforward class name in a string which is not scoped. This is the first problem encountered. When fixed, you find another when Pretty Printing. I assume at that point we've gone through most of the parsing, so there may just be a couple of these.

`PhpParser\PrettyPrinterAbstract::p()`:
```php
/**
 * Pretty prints a node.
 *
 * @param Node $node Node to be pretty printed
 *
 * @return string Pretty printed node
 */
protected function p(Node $node)
{
    return $this->{'p' . $node->getType()}($node);
}
```
The `$node->getType()` returns the class ref which is prefixed, so the function name is corrupted and causes a fatal error.

The outcome of the problem is that scoped files are empty apart from an opening PHP tag. So essentially, we need a facility to apply simple patches to files. I'm not sure if they should target before or after scoping. I suspect doing it beforehand is simpler, but debugging takes place after (and pretty printing meddles with line numbers in between). Might be useful to have a shortcut to extract the phar from php-scoper itself for ease of debugging. I'm using the following the extract the phar at the moment:

`php -r '$phar = new Phar("bin/php-scoper.phar"); $phar->extractTo("./build/extract");'`

theofidry:
I think we would be better off trying to patch PhpParser rather

padraic:
Just for reference, the second example can be resolved with (23 being uniqid() string length):

`PhpParser\NodeAbstract::getType()`:
```php
public function getType()
{
    return strtr(substr(rtrim(get_class($this), '_'), 15+23), '\\', '_');
}
```
No further errors thereafter.

As regards patching, we can patch this directly (if that's what you mean, via makefile), but the prefix is not static so we still need a way to advise the patching op of what the prefix is, so patching needs to be done either during scoping (which the issue title suggests) or post-scoping when we can signal or retrieve it - perhaps even just dump the prefix to a simple file for reading instead of searching for it.

theofidry:
[...]

PR #65, opened 2017-06-23
Add StringReplacer scoper and config for build

padraic:
Possible implementation of #63 for feedback using `preg_replace`. Works to current build with the scoper.json configuration (if you can survive the nested slash escapes in there!).


src/Console/Command/AddPrefixCommand.php
```diff
@@ -234,4 +251,76 @@ private function validateOutputDir(InputInterface $input, OutputStyle $io)
...
+        $replaceStrings = $input->getOption(self::REPLACE_STRINGS_OPT);
+
+        if (false === $this->fileSystem->isAbsolutePath($replaceStrings)) {
+            $replaceStrings = getcwd().DIRECTORY_SEPARATOR.$replaceStrings;
+        }
+
+        $input->setOption(self::REPLACE_STRINGS_OPT, $replaceStrings);
+
+        if (false === $this->fileSystem->exists($replaceStrings)) {
+            return;
```
padraic: Should raise an exception actually...

padraic:
Not sure why Scrutinizer is erroring...

scoper.json
```diff
@@ -0,0 +1,18 @@
+{
```
theofidry: Maybe we should go for https://github.com/nette/neon? Unlike JSON, multiline is either which could help making the code a bit clearer. Alternatively we could allow a PHP config.

To be clear I don't have anything against JSON and it's a popular config format, but in our case it looks really awkward to use

src/Console/Command/AddPrefixCommand.php
```diff
@@ -98,6 +109,12 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $this->validatePaths($input);
         $this->validateOutputDir($input, $io);
 
+        $replaceStrings = $input->getOption(self::REPLACE_STRINGS_OPT);
+        if (is_string($replaceStrings)) {
```
theofidry: `null !== replaceStrings `?

src/Scoper/StringReplacer.php
```diff
@@ -0,0 +1,89 @@
...
+     * Replaces selected strings in files if configured by user.
+     *
+     * {@inheritdoc}
+     */
+    public function scope(string $filePath, string $prefix): string
+    {
+        $content = $this->decoratedScoper->scope($filePath, $prefix);
+
+        if (is_array($this->replaceMap) && isset($this->replaceMap[$filePath])) {
+            foreach ($this->replaceMap[$filePath] as $replacement) {
```
theofidry: could do an early return here

src/Scoper/StringReplacer.php
```diff
@@ -0,0 +1,89 @@
...
+                        )
+                    );
+                }
+            }
+        }
+
+        return $content;
+    }
+
+    private function mapByFile(array $replaceMap)
```
theofidry: can we make it stateless? i.e. returning `replaceMap` instead of mutating the property directly

src/Scoper/StringReplacer.php
```diff
@@ -0,0 +1,89 @@
...
+
+use Humbug\PhpScoper\Scoper;
+use Humbug\PhpScoper\Throwable\Exception\RuntimeException;
+
+final class StringReplacer implements Scoper
+{
+    /**
+     * @var array
+     */
+    private $replaceMap;
```
theofidry: the default value could be an empty array to avoid the [`is_array($this->replaceMap`](https://github.com/humbug/php-scoper/pull/65/files#diff-116fe39c87c6067b54d0c5cb7b90f772R48) and remove the `@var` (as the type would become clear from the declaration)

src/Scoper/StringReplacer.php
```diff
@@ -0,0 +1,89 @@
...
+
+    private $decoratedScoper;
+
+    public function __construct(Scoper $decoratedScoper)
+    {
+        $this->decoratedScoper = $decoratedScoper;
+    }
+
+    public function setReplaceMap(array $replaceMap)
```
theofidry: `configureReplaceMap()`?

Makefile
```diff
@@ -31,7 +31,7 @@ build: vendor
 	# be used and `box.json.dist` must include the `tests` directory
 	#
 	composer install --prefer-dist
-	php -d zend.enable_gc=0 bin/php-scoper.php add-prefix --force
+	php -d zend.enable_gc=0 bin/php-scoper.php add-prefix --replace-strings=scoper.json --force
```
theofidry: should we add a default? I.e. detect `scoper.json` when it exists

src/Console/Command/AddPrefixCommand.php
```diff
@@ -37,19 +38,23 @@
...
 
     private $fileSystem;
     private $handle;
+    private $stringReplacer;
 
     /**
      * @inheritdoc
      */
-    public function __construct(Filesystem $fileSystem, HandleAddPrefix $handle)
+    public function __construct(Filesystem $fileSystem, HandleAddPrefix $handle, StringReplacer $stringReplacer = null)
```
theofidry: I'm not a big fan of this pattern, it's really not clear that we are changing the state of `Scoper` which is used by the handler that way, besides making the scoper stateful. What about passing the `replaceMap` as a parameter or the handler and scoper?

scoper.json
padraic: Yes, JSON needs to go - the nested escaping is ridiculous in this case.

src/Console/Command/AddPrefixCommand.php
padraic: I assume you’re referring to a ReplaceMap object above? s/or/for/ ;).

theofidry: `public function setReplaceMap(array $replaceMap)` => instead of using a mutator that way, I would just pass it to the handler. Not sure if we should encapsulate that into a Value Object to make it clearer, but IMO that would be cleaner and less confusing to pass it as a parameter

padraic:
@theofidry Switched to YAML. The one thing I'm blocking on is the Scoper redirection to set the replacement strings. Every option I arrive at looks equally bad, and it boils down to either living with it, extracting some of the input processing to get the map early enough, or possible overkill for something this small, encapsulating object creation for deferral to when we have the replacements.

Updated some of the tests, though there's none specific to the StringReplacer for now.

src/Console/Command/AddPrefixCommand.php
```diff
@@ -234,4 +253,96 @@ private function validateOutputDir(InputInterface $input, OutputStyle $io)
...
+                    'Unable to parse replacement strings JSON file: %s%s%s',
+                    $replaceStrings,
+                    PHP_EOL,
+                    $e->getMessage()
+                )
+            );
+        }
+    }
+
+    private function makeAbsolutePath(string $path): string
```
padraic: Ye olde extract method, should be propagated to a few methods in this class where this check is repeated.

theofidry:
@padraic I'm a bit worried about that approach, it does work in this specific case but looks quite limited and brittle. I would be fine if it was internal, but that's an important extension point. So thinking a bit more about here he's another approach I would like to suggest:

```php
// .php-scoper.php
<?php

return [
    function(string $filePath, string $prefix): string {
        // Closure actually matching the scoper signature
    },
];
```

There is two differences here:

- It gives greater flexibility in the string manipulation which allow to go beyond the limitations of `preg_replace()`
- As it's pure PHP, you can do more hacky things

Although I'm usually against the second point, here's a reason why I would allow it in our case. For example you have a library for which one of the dependency is not a fixed version but a ranged one, i.e. not fixed by a `composer.lock`. So let's take the example of `humbug` which could depend on `nikic/php-parser`. On `3.0.5`, we need to apply a patch with the extension provided, however in `3.0.6` this patch is no longer valid (we are manipulating potentially internal code so this kind of BC can happen in a patch version). So what we want to do here is something like:

```php
// .php-scoper.php
<?php

require_once __DIR__.'/vendor/ocramius/package-versions/src/PackageVersions/Versions.php';

$patchPhpParser = function(string $filePath, string $prefix): string
{
    $version = explode('@', Versions::getVersion('nikic/php-parser'))[0];

    if ('v3.0.5' === version) {
       // do something
    }

    return file_get_contents($filePath);
}

return [
    patchPhpParser,
];
```

WDYT?

padraic:
@theofidry I'm not worried about brittleness per se. I think that updates over time may be inevitable, but we're likely to be building PHARs at single points in time with known dependencies. There could be a use case where PHP-Scoper is needed, and PHARs are built to many different dependencies, and where a compiled list of changes by version have value. I just don't know of one personally off hand.

However, looking at JSON and then YAML, and writing the changes into both formats, removing the extra layer and just using PHP is attractive all by its lonesome, so I think it's the better simpler overall solution for everyone's sanity. That it can help with various other scenarios are bonus points.

theofidry:
My example may not be the best, but overall I'm just a bit worried that the JSON/YAML makes it overly cumbersome and hard to debug, whereas the approach I suggested is much more flexible and the user can do whatever he wants to apply a patch. It certainly opens a door to fuck up, but the audience we are targeting is likely to be quite advanced skill wise and with enough with a proper doc I think we won't have too much complains from it

theofidry:
Closing in favour of #67 then :)

PR #67, opened 2017-06-29
Add patchers

theofidry:
Attempt to replace https://github.com/humbug/php-scoper/pull/65. It's still not finished as:

- It doesn't really work yet (the e2e tests still fails)
- Tests are missing
- Maybe something else I forgot

Closes https://github.com/humbug/php-scoper/issues/63

@padraic if you're happy with the idea I'll try to finish that PR (unless you want to finish it ofc).


padraic:
@theofidry Fire away and finish! :)

theofidry:
@padraic were the patches that you applied enough? The e2e tests are still failing for me so maybe I got them wrong...

padraic:
@theofidry Hmm, it just needed those two patches. It could just be the makefile - if you check my PR, I made some changes in there to get everything passing: https://github.com/humbug/php-scoper/pull/65/files#diff-b67911656ef5d18c4ae36cb6741b7965L64

theofidry:
Will take a look

theofidry:
Ok added some tests. I'm still a bit unhappy with the name as it's the same as the binary one which can be a bit confusing...

Also regarding the error I could not find the source, however it looks like the regex doesn't anything (the code we are looking for doesn't exists anymore). I suppose this may be due to having different versions of PHP-Parser... Given how tightly coupled to it we are, we should fix it to a patch version and/or publish the `composer.lock` version.

I think publish the `.lock` version makes sense: it's mandatory if we want to publish php-scoper as a PHAR, i.e. treat it as an application. And when required with Composer, no patch is needed to make it work.

padraic:
PHP-Parser 3.0.6 was released about a week back. Checking it, I think they updated one line we are seeking to replace (to remove it and use single backslashes): https://github.com/nikic/PHP-Parser/blob/0808939f81c1347a3c8a82a5925385a08074b0f1/lib/PhpParser/Lexer.php#L363

Updated the regular expression used ought to fix that as an issue.

theofidry:
@Nyholm any suggestions for the config file name?

Nyholm:
Lets keep it simple: `scoper.ext` or `php-scoper.ext`.

theofidry:
`.ext`? I would rather keep the `.php`, but `scoper.php` sounds better 👍

padraic:
We could always return to the convention for include files: `scoper.inc`

PR #74, opened 2017-07-05
Add support for class_exists('string')

### LINKED ISSUE #73: Fix class_exists
This is not rewritten: 

```php
if (!class_exists('Symfony\\Component\\Yaml\\Parser')) {
     throw new RuntimeException('Unable to load YAML config files as the Symfony Yaml Component is not installed.');
}
```

Nyholm:
We should support `class_exists('\\Symfony\\Component\\Yaml\\Yaml')`. 

(Im having some issues running this lib over Symfony code base.)

This will fix #73

### DIFF
--- a/src/NodeVisitor/FunctionCallScoperNodeVisitor.php
+++ b/src/NodeVisitor/FunctionCallScoperNodeVisitor.php
@@ -17,16 +17,23 @@ namespace Humbug\PhpScoper\NodeVisitor;
 use PhpParser\Node;
 use PhpParser\NodeVisitorAbstract;
 
-final class ClassExistsScoperNodeVisitor extends NodeVisitorAbstract
+final class FunctionCallScoperNodeVisitor extends NodeVisitorAbstract
 {
     /**
      * @var string
      */
     private $prefix;
 
-    public function __construct($prefix)
+    /**
+     * Function which first parameter should be prefixed
+     * @var array
+     */
+    private $functions;
+
+    public function __construct($prefix, array $functions = ['class_exists', 'interface_exists'])
     {
         $this->prefix = $prefix;
+        $this->functions = $functions;
     }
 
     /**
@@ -42,7 +49,8 @@ final class ClassExistsScoperNodeVisitor extends NodeVisitorAbstract
             return $node;
         }
 
-        if ('class_exists' !== $node->name->getFirst()) {
+
+        if (!in_array($node->name->getFirst(), $this->functions)) {
             return $node;
         }
 
@@ -51,7 +59,9 @@ final class ClassExistsScoperNodeVisitor extends NodeVisitorAbstract
             return $node;
         }
 
-        $value->value = $this->prefix.'\\'.$value->value;
+        if ($this->prefix !== $value->value) {
+            $value->value = $this->prefix.'\\'.$value->value;
+        }
 
         return $node;
     }
--- a/src/NodeVisitor/FunctionCallScoperNodeVisitor.php
+++ b/src/NodeVisitor/FunctionCallScoperNodeVisitor.php
@@ -25,7 +25,8 @@ final class FunctionCallScoperNodeVisitor extends NodeVisitorAbstract
     private $prefix;
 
     /**
-     * Function which first parameter should be prefixed
+     * Function which first parameter should be prefixed.
+     *
      * @var array
      */
     private $functions;
@@ -48,7 +49,6 @@ final class FunctionCallScoperNodeVisitor extends NodeVisitorAbstract
         if (!$node->name instanceof Node\Name) {
             return $node;
         }
-
 
         if (!in_array($node->name->getFirst(), $this->functions)) {
             return $node;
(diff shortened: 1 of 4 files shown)


theofidry:
@padraic well @Nyholm doesn't restrict to `class_exists()` as the tests shows.

Regarding 1. it's a fair point, although it might be weird one can do `$x = 'Closure` or `$x = '\Closure'`

padraic:
@theofidry I'm only seeing `class_exists` in the test change restricted to literal strings/::class constant?

Re-reading my feedback while here, @Nyholm can note I'm fine with this being a narrow focused PR for `class_exists` as the most common usage. The last bit was something we can expand on in later PRs.

Nyholm:
I've updated the PR. 

> The string is inspected and we skip prefixing non-namespaced classes (ignored as a default behaviour elsewhere).

Im not sure I understand this. How can I do this?

> Re-reading my feedback while here, @Nyholm can note I'm fine with this being a narrow focused PR for class_exists as the most common usage. The last bit was something we can expand on in later PRs.

Im happy with the feedback. I do not mind widening the use cases a little

tests/Scoper/PhpScoperTest.php
```diff
@@ -965,5 +965,114 @@ function foo($bar) : Closure
...
+
+class_exists('Humbug');
+
+PHP
+            ,
+            'Humbug',
+            <<<'PHP'
+<?php
+
+class_exists('Humbug');
```
theofidry: this one may look a bit weird because the class in question is the same as the prefix. Just for the sake of clarity I would pick another name `Foo` or `PhpScoper`

padraic:
@Nyholm Sorry, I was referring to something like `Closure` vs `Namespace\SomeClass` - we default to prefixing the second, but ignoring the first from global namespace.

tests/Scoper/PhpScoperTest.php
```diff
@@ -965,5 +965,118 @@ function foo($bar) : Closure
...
+class_exists('\\Symfony\\Component\\Yaml\\Yaml');
+
+PHP
+            ,
+            'Humbug',
+            <<<'PHP'
+<?php
+
+class_exists('Humbug\\Symfony\\Component\\Yaml\\Yaml');
+class_exists('Humbug\\Symfony\\Component\\Yaml\\Yaml');
```
Nyholm: Is this correct? Or should it be `'\\Humbug\\Symfony\\Component\\Yaml\\Yaml'`

Nyholm:
Thank you for the reviews: 

> Also unless I'm mistaken, the current changes apply to any function right?

You **could** apply it to any function. See constructor parameter.

src/NodeVisitor/FunctionCallScoperNodeVisitor.php
```diff
@@ -0,0 +1,71 @@
...
+        if (!$node->name instanceof Node\Name) {
+            return $node;
+        }
+
+        if (!in_array($node->name->getFirst(), $this->functions)) {
+            return $node;
+        }
+
+        $value = $node->args[0]->value;
+        if (!$value instanceof Node\Scalar\String_) {
```
Nyholm: @theofidry This makes sure the first argument to the function is a plain string. If anything else (like concat or variables) it will be ignored.

theofidry: What happens if the next one is not?

theofidry: Actually nvm, it should be ok if it's not, I would just add a test for it still

src/NodeVisitor/FunctionCallScoperNodeVisitor.php
```diff
@@ -0,0 +1,71 @@
...
+    private $prefix;
+
+    /**
+     * Function which first parameter should be prefixed.
+     *
+     * @var array
+     */
+    private $functions;
+
+    public function __construct($prefix, array $functions = ['class_exists', 'interface_exists'])
```
theofidry:
```php
/**
 * @param string   $prefix
 * @param string[] $functions
 */
public function __construct(string $prefix, array $functions)
```

and pass the default values upstream as it's a config thing really

src/NodeVisitor/FunctionCallScoperNodeVisitor.php
```diff
@@ -0,0 +1,71 @@
...
+{
+    /**
+     * @var string
+     */
+    private $prefix;
+
+    /**
+     * Function which first parameter should be prefixed.
+     *
+     * @var array
```
theofidry: `string[]`, but to be fair, for both `$prefix` and `$functions`, as they are defined just bellow with the typehint I wouldn't add the phpdoc here:

```php
private $prefix;
private $functions;
```

tests/Scoper/PhpScoperTest.php
theofidry: eeh, I would say both are correct so for the sake of consistency we could transform it to `'\\Humbug\\Symfony\\Component\\Yaml\\Yaml'` to be clearer

src/NodeVisitor/FunctionCallScoperNodeVisitor.php
Nyholm: There is a test for concatenated strings already =)

Nyholm:
Thank you for the review. I've updated the PR.

theofidry:
Thank you @Nyholm :)

Nyholm:
Thank you for the reviews!

PR #75, opened 2017-07-06
Fix patcher

theofidry:
- Add `composer.lock`
- Remove Composer patch
- Update Makefile rules
- Fix the patcher

@padraic looks like this patcher is not enough but not really sure how that needs to be fixed. Might have a look over the weekend...


padraic:
It doesn't look the makefile changes I had made it across. The problem I had, after patching, is that set005 references a package in composer.json, but we don't install it via composer, and the fixture doesn't include the vendor directory. So I just copied the fixture to build, `composer install`, scoped back to the expected build directory, and that was it.

I assume it's failing due to not found namespace for use statement here: https://github.com/theofidry/php-scoper/blob/0f833385e89c27219439c3f4b5ac2724c120c482/fixtures/set005/bin/greet.php

theofidry:
Hm indeed this could be a problem^^, we should install it via Composer for sure

theofidry:
Hm Same issue still, actually note that this is happening for the `set004`

theofidry:
Does it still make sense to test lowest deps? We are now adding a `composer.lock`...

PR #76, opened 2017-07-08
Scope whitelisted global namespace classes

### LINKED ISSUE #10: Add namespace to non-namespaced classes
Currently the classes without namespace (for example https://github.com/sebastianbergmann/phpunit/blob/master/src/Framework/TestCase.php) are just ignored

theofidry:
Closes #10

### DIFF
--- a/src/NodeVisitor/GlobalWhitelistedNamesNodeVisitor.php
+++ b/src/NodeVisitor/GlobalWhitelistedNamesNodeVisitor.php
@@ -14,19 +14,22 @@ declare(strict_types=1);
 
 namespace Humbug\PhpScoper\NodeVisitor;
 
+use Closure;
 use PhpParser\Node;
 use PhpParser\Node\Name;
 use PhpParser\Node\Stmt\GroupUse;
 use PhpParser\Node\Stmt\UseUse;
 use PhpParser\NodeVisitorAbstract;
 
-final class UseNamespaceScoperNodeVisitor extends NodeVisitorAbstract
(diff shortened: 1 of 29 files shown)


src/NodeVisitor/GlobalWhitelistedNamesNodeVisitor.php
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function enterNode(Node $node)
+    {
+        if ($node instanceof Name
+            && 1 === count($node->parts)
+            && true === ($this->whitelister)($node->parts[0])
```
atimmer: In the `Name` class PHPParser writes:

> Avoid directly accessing $parts, use methods instead.

So I think `$node->parts[0]` should become `$node->getFirst()`

atimmer:
I'm testing this and it seems that it doesn't prefix the name in the `class SomeClass` definition. It looks like the name of a `Class_` isn't actually a `Name` object, but just a string.

theofidry:
Thanks for testing :) Could you provide a more complete example?

src/NodeVisitor/GlobalWhitelistedNamesNodeVisitor.php
padraic: @atimmer I cam across this before, but the current status is that there is no public API for parts. Once it's added, we can move across. The deprecation seems more of a TODO in this case.

theofidry:
PR updated, ready for review

PR #78, opened 2017-07-09
Add stop on failure option

### LINKED ISSUE #71: Error when symfony/phpunit-brige is installed
This is a minor issue. I get the following error
> Argument 1 passed to Humbug\PhpScoper\Scoper\Composer\AutoloadPrefixer::prefixPackageAutoloads() must be of the type array, string given, called in /Users/tobias/.composer/vendor/humbug/php-scoper/src/Scoper/Composer/InstalledPackagesScoper.php on line 75   

Because of https://github.com/symfony/phpunit-bridge/blob/3.4/Tests/DeprecationErrorHandler/fake_vendor/composer/installed.json

theofidry:
Closes #71


theofidry:
Yeah it's for sure going to be annoying... Do you have any suggestions on how to do it? I still want to a test to ensure we have a stack trace of some kind

padraic:
I'd do a few very general assertions, essentially defining what the stack trace is but without being too specific, e.g. phrases `stack phrase`, `{main}`, `TypeError`, >N lines (but less than currently to allow for file loss in refactoring upstream).

atimmer:
This doesn't actually close #10, right?

theofidry:
Right that was #71, soz about that

PR #82, opened 2017-07-11
Fix AppVeyor builds

theofidry:


theofidry:
@padraic could you give me access to AppVeyor so that I could interrupt AppVeyor builds when necessary?

padraic:
@theofidry Can you check now? Added @humbug/core members as a role to PHP Scoper there.

theofidry:
Got it. If someone of the @humbug/core knows how to set up Make for Windows... that would be helpful...

I don't have a Windows machine at all so trying to do it and waiting on AppVeyor each time is a real pain as AppVeyor is super slow

padraic:
I was really hoping not to have to boot up Windows for anything other than Subnautica today ;).

theofidry:
Fortunately it just worked, that's a good game btw :)

PR #87, opened 2017-07-15
[WIP] Update readme

padraic:
@theofidry Updated the Readme in bits and pieces over the week, so given #86 might as well stash the draft to a PR for feedback.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,42 +8,36 @@
 [![Slack](https://img.shields.io/badge/slack-%23humbug-red.svg?style=flat-square)](https://symfony.com/slack-invite)
 [![License](https://img.shields.io/badge/license-MIT-red.svg?style=flat-square)](LICENSE)
 
-PHP-Scoper is a tool for adding a prefix to all PHP namespaces in a given file
-or directory. 
-
+PHP-Scoper is a tool which essentially moves any body of code, including all
+dependencies such as vendor directories, to a new and distinct namespace.
 
 ## Goal
 
-PHP-Scoper's goal is to make sure that all code in a directory lies in a 
-distinct PHP namespace. This is necessary when building PHARs that 
+PHP-Scoper's goal is to make sure that all code for a project lies in a 
+distinct PHP namespace. This is necessary, for example, when building PHARs that:
 
-* Bundle their own vendor dependencies
-* Load code of arbitrary PHP projects
+* Bundle their own vendor dependencies; and
+* Load/execute code from arbitrary PHP projects with similar dependencies
 
 These PHARs run the risk of raising conflicts between their bundled vendors and 
-the vendors of the loaded project, if the vendors are required in incompatible
-versions.
-
+the vendors of the project it is interacting with, where the PHAR's dependencies
+are used preferentially leading to difficult to debug issues due to dissimilar or
+unsupported package versions.
 
 ## Installation
 
-You can install PHP-Scoper with Composer:
+The preferred method of installation is to use the PHP-Scoper phar, which can
+be downloaded from the most recent Github Release. Subsequent updates can be
+downloaded by running:
 
 ```bash
-composer global require humbug/php-scoper:dev-master
+php-scoper.phar self-update
 ```
 
-If you cannot install it because of a dependency conflict or you prefer to
-install it for your project, we recommend you to take a look at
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -8,72 +8,228 @@
...
-* Load code of arbitrary PHP projects
+* Bundle their own vendor dependencies; and
+* Load/execute code from arbitrary PHP projects with similar dependencies
 
 These PHARs run the risk of raising conflicts between their bundled vendors and 
-the vendors of the loaded project, if the vendors are required in incompatible
-versions.
-
+the vendors of the project it is interacting with, where the PHAR's dependencies
+are used preferentially leading to difficult to debug issues due to dissimilar or
```
theofidry: Suggestion:

```md
When a package (of possibly different versions) exists is found in both a PHAR
and the executed code, the one from the PHAR will be used. This means these
PHARs run the risk of raising conflicts between their bundled vendors and the
vendor of the project they are interacting with, leading to issues potentially
very difficult to debug due to dissimilar or unsupported package versions.
```

README.md
```diff
@@ -8,72 +8,228 @@
...
+unsupported package versions.
 
 ## Installation
 
-You can install PHP-Scoper with Composer:
+The preferred method of installation is to use the PHP-Scoper phar, which can
```
theofidry: `PHAR`

theofidry: maybe we should add a link to the releases page?

README.md
```diff
@@ -8,72 +8,228 @@
...
+be downloaded from the most recent Github Release. Subsequent updates can be
+downloaded by running:
 
 ```bash
-composer global require humbug/php-scoper:dev-master
+php-scoper.phar self-update
 ```
 
-If you cannot install it because of a dependency conflict or you prefer to
```
theofidry: I would divide the installation in two parts: "PHAR (preferred)" and "Composer". Hence keeping the paragraph below still just in case. We could remove the `:dev-master` part though

README.md
```diff
@@ -8,72 +8,228 @@
...
+
+This and similar issues *may* arise after scoping, and can be debugged by
+running the scoped code and checking for issues. For this purpose, having a
+couple of end to end tests to validate post-scoped code or PHARs is recommended.
+
+Applying such a change can be achieved by defining a suitable patcher in
+`scoper.inc.php`:
+
+```php
+return [
```
theofidry: I would add:

```php
<?php declare(strict_types=1)

// scoper.inc.php

return [
    // ....
];
```

to make it clear that's all the file needs

README.md
```diff
@@ -8,72 +8,228 @@
...
+non-global. In other words, non-namespaced code is not scoped. This leaves the
+majority of classes, functions and constants in PHP, and most extensions,
+untouched.
+
+This is not necessarily a desireable outcome for vendor dependencies which are
+also not namespaced. To ensure they are isolated, you can configure PHP-Scoper to
+allow their prefixing from `scoper.inc.php` using basic strings or callables:
+
+```php
+return [
```
theofidry: same as above (careful you're missing an ending semi-colon as well)

README.md
```diff
@@ -8,72 +8,228 @@
...
+## Building A Scoped PHAR
+
+This is a brief run through of the basic steps encoded in PHP-Scoper's own
+[Makefile](Makefile) and elsewhere to build a PHAR from scoped code.
+
+### Step 1: Configure build location and prep vendors
+
+If, for example, you are using [Box](box) to build your PHAR, you
+should set the `base-path` configuration option in your `box.json` file
+to point at the 'build' directory which will host scoped code.
```
theofidry: Maybe add "(which is `build` by default)" at the end

README.md
```diff
@@ -8,72 +8,228 @@
...
+
+```js
```
theofidry: js?

README.md
```diff
@@ -8,72 +8,228 @@
...
+```bash
+bin/php-scoper add-prefix
+```
+
+As there are no path arguments, the current working directory will be scoped to
+`./build` in its entirety. Of course, actual prefixing is limited to PHP files,
+or PHP scripts. Other files are copied unchanged, though we also need to scope
+certain Composer related files.
+
+Speaking of scoping Composer related files...the next step is to dump the
```
theofidry: `files... The`

PR #88, opened 2017-07-17
Allow to whitelist some code

theofidry:
The whole point of scoping is to avoid any conflict between the shared code between the code bundled in the PHAR and the consumer code. However some parts must be exempt from that rule. For example with PHPUnit, the `PHPUnit\Framework\TestCase` class must _not_ be scoped.


theofidry:
@padraic heads up.

So after a bit of tinkering, there is an issue remaining with my current approach which consist of not scoping certain classes: the autoloading. To explain further, I'll take the `set011` as a base example.

That example is rather simple and try to imitate what some tools like PHPUnit does:

- A `greet.php` binary which tries to locate `Set011\Dictionaries` classes located in the `tests` directory
- The `Set011\Dictionary` interface defined in `src` with other classes
- The `Set011\SalutationDictionary` defined in `tests` which implements `Set011\Dictionary`
- The scoping scopes everything except `Set011\Dictionary` and the classes defined in `tests`

What we want: the scoping should be smart enough to ensure in the generated PHAR, the usages of the class `Set011\Dictionary` are not prefixed, hence a user can consume this class to write his code independently of the PHAR: `Set011\Dictionary` is a shared class between the bundled code and the consumer code.

What currently happens with my implementation: well first of `Set011\Dictionary` is not found as is in a different namespace although in an autoloaded location. This is where there is two ways to go.

<hr>

### First way

A lazy way but that would hurt the usability of PHP-Scoper: edit the `composer.json` generated to add back an entry in the according autoloading section and require to change all the usages of `Set011\Dictionary` to a FQCN.

So here for example we had:

```json
{
    "bin": [
        "bin/greet"
    ],
    "autoload": {
        "psr-4": {
            "Set011\\": "src/"
        }
    },
    "autoload-dev": {
        "psr-4": {
            "Set011\\": "tests/"
        }
    }
}
```

Assuming the prefixed applied is `PhpScoper598627d709dd4`, we need to edit the generated `composer.json` to add the missing autoloading entry:

```json
{
    "bin": [
        "bin/greet"
    ],
    "autoload": {
        "psr-4": {
            "PhpScoper598627d709dd4\\Set011\\": "src/"
            "Set011\\": "src/"
        }
    },
    "autoload-dev": {
        "psr-4": {
            "Set011\\": "tests/"
        }
    }
}
```

And then if you have a piece of code like:

```php
 namespace Set011;

use Dictionary;

// ...
if ($x instanceof Dictionary) {
    // ...
}
```

or:

```php
 namespace X;

use Set011\Dictionary;

// ...
if ($x instanceof Dictionary) {
    // ...
}
```

it needs to be changed to:

```php
 namespace Set011;

// ...

if ($x instanceof \Set011\Dictionary) {
    // ...
}
```

and:

```php
 namespace X;

// ...
if ($x instanceof \Set011\Dictionary) {
    // ...
}
```

The leading backslash being mandatory as this is what allows Composer to include the right file. If this is missing he will try to guess the file location relying on the autoloading convention which is likely to fail.

<hr>

### Second way

Be smarter. We have the AST tree, we can work on trying to do that FQCN transformation on our side. It does involves more work on our end but that's doable.

I'm actually not too worried about the transformation, as I said we have the AST tree so that's doable even if performant can suffer from that. Other worrying points are:

- The feasibility to update the `composer.json` automatically
- How to deal with cases where multiple paths are passed to edit the right `composer.json`

In any cases, the whitelist has one clear limitation: it only works for user code, not vendor code.

Any thoughts?

tests/Scoper/PhpScoperTest.php
```diff
@@ -918,7 +1362,11 @@ public function provideValidFiles()
 PHP
         ];
 
-        yield '[Single part global namespace reference] a full qualified class reference' => [
+        //TODO: do not allow to whitelist a class from the global namespace, e.g. `Foo`. Indeed they are already
```
theofidry: pretty important TODO

theofidry:
Update: I added more tests and I think we are done on that side (for PhpScoperTest). What's left is make those tests pass and then do a bit of cleanup before the review. That said if someone wants to review PhpScoperTest that can already be done right now

tests/Scoper/files_to_scope/class-const/global-scope-single-part-with-single-level-use-statement-and-alias.php
```diff
@@ -0,0 +1,86 @@
+<?php
+
+declare(strict_types=1);
```
theofidry: move the specifications elsewhere

PR #89, opened 2017-07-20
Add finders

### LINKED ISSUE #84: Pass custom finders
I think we should provide a file discovery experience similar to box to ease the usage outside of box (and we don't know when the box integration will be done) which is easily doable by adding 2 entries in the config.

@padraic thoughts?

theofidry:
Closes #84.


theofidry:
Not sure why Scrutinizer is failing, it might be easier to just upload the clover.xml rather than running their CI which I guess may require its own config.

I also don't have access to the Scrutinizer config but we should exclude `fixtures` from the inspection.

padraic:
I've updated the Scrutinizer config.

theofidry:
I find the Symfony Finder exclude methods not very easy to use and helpful. For example:

```php
$finder = (new Finder())
    ->files()
    ->in(__DIR__)
    ->exclude(__DIR__.__DIRECTORY_SEPARATOR__.'tests')
;
```

with:

```
bin/
    foo
src/
    Foz.php
    Bar.php
tests/
    FozTest.php
composer.json
```

will not exclude `tests/FozTest.php`. So I think I'll add a `blacklist-finders` entry as well

src/Console/Configuration.php
```diff
@@ -85,13 +116,67 @@ public function getGlobalNamespaceWhitelisters()
...
+        if (false === array_key_exists(self::FINDER_KEYWORD, $config)) {
+            return [];
+        }
+
+        $finders = $config[self::FINDER_KEYWORD];
+
+        if (false === is_array($finders)) {
+            throw new InvalidArgumentException(
+                sprintf(
+                    'Expected finders to be an array of callables, found "%s" instead.',
```
pamil: We can leave the `of callables` part as it isn't asserted here.

theofidry:
ping @padraic

PR #92, opened 2017-08-30
Add the specifications

theofidry:
Extracted the tests refactoring from #88 here. This PR is only about refactoring/adding the tests and not implementing them which means the build failure is expected.


theofidry:
Another idea that came whilst writing the tests and talking with Ocramius: we could do the prefixing in two passes and during the first pass, try to resolve all the references to transform calls into FQ calls. The prefixing would then be a lot simpler and the annoying cases like:

```php
<?php

use Foo\Bar;

Bar\X::MAIN_CONST;
new Bar();
```

becomes easier to resonate about

theofidry:
@Ocramius ready for review I think.

Maybe it requires a bit of context though:

- the tool is about isolating code, for example for building PHARs for avoiding any collision between the bundled code and executed code
- there is two concepts of "whitelist":
  - whitelist for global classes: classes belonging to the global namespace like `Closure` are not prefixed. However some tools/frameworks, e.g. Symfony with `AppKernel` or PHPUnit still put some classes in the global namespace and as such, whitelisted classes are properly prefixed
  - whitelist for a regular namespaced classe: to avoid to prefix them. This will be used to allow to share code between the isolated PHAR and the executed code like the test case class for PHPUnit.

Regarding this PR:

- This is only the specs, the implementation following it will follow later in other PRs
- The objective is to allow this tool to work well with most of the code but within reasonable limits: shitty completely dynamic code is not supported. There is some extension point to allow pretty dirty stuff, but we won't go further than that.
- I'm pinging you as I know you are familiar with that kind of work (dirty hacky tools) so in case you might think of a tricky case that we might want to support

Also that's a lot of specs, a thorough review is welcomed but by no means feel obliged to review everything. A quick glance is perfectly fine

theofidry:
Going to merge that one to try to make some progress on the implementation, reviews are still welcome though

PR #94, opened 2017-09-05
Implement the specifications

theofidry:
Make the unit tests green again.


theofidry:
The implementation is done and the tests should pass. Now what is left is cleaning up and refactoring.

PR #110, opened 2017-09-30
Enable e2e tests

theofidry:


theofidry:
It's working locally, so not sure why it's broken on Travis... @padraic could you try it on your end see if that works? (just run `make test`)

PR #117, opened 2017-10-07
Use BetterReflection for global classes

theofidry:
Instead of not touching the global classes at all and prefix them only when explicitly specified, we can prefix them if they are not internal. We could then get rid of the global whitelist and allow the whitelist for classes belonging to the global namespace.
  

- [x] Remove the global namespace config concept
- [x] Simplify the code (now possible as there is only 1 whitelist concept)
- [x] Allow whitelist to work with classes from the global namespace
- [x] Update the tests
- [x] Update the configuration

Current status: did a replacement and updating the tests; need more refactoring and checking the list above


TODO-roave
```diff
@@ -0,0 +1,112 @@
+// vendor/roave/better-reflection/src/SourceLocator/Ast/Locator.php
```
theofidry: TODO: this should be a patch for `roave/better-reflection`

phpunit.xml.dist
```diff
@@ -10,7 +10,7 @@ file that was distributed with this source code.
 -->
 
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/6.0/phpunit.xsd"
+         xsi:noNamespaceSchemaLocation="vendor/phpunit/phpunit/phpunit.xsd"
```
theofidry: this can be part of another PR

tests/Scoper/PhpScoperTest.php
```diff
@@ -396,11 +418,48 @@ public function test_can_scope_valid_files(string $spec, string $contents, strin
 
         $patchers = [create_fake_patcher()];
 
-        $whitelister = function (string $className) {
-            return 'AppKernel' === $className;
-        };
+        $nonInternalReflectionProphecy = $this->prophesize(ReflectionClass::class);
```
theofidry: no longer necessary

src/NodeVisitor/NamespaceStmtPrefixer.php
```diff
@@ -146,13 +142,15 @@ private function isWhitelistedNode(Node $node)
...
+        return null === $namespace->name || (null !== $namespace->name && $this->prefix !== $namespace->name->getFirst());
+
         if (null !== $namespace->name && $this->prefix !== $namespace->name->getFirst()) {
             return true;
         }
 
-        if (null === $namespace->name && $this->hasWhitelistedNode([$namespace])) {
-            return true;
-        }
+//        if (null === $namespace->name && $this->hasWhitelistedNode([$namespace])) {
```
theofidry: can be simplified

src/NodeVisitor/NamespaceStmtPrefixer.php
```diff
@@ -41,47 +41,45 @@
...
-     */
-    public function leaveNode(Node $node)
-    {
-        return (
-            !$this->hasWhitelistedNode
-            || $node instanceof Namespace_
-            || AppendParentNode::hasParent($node)
-        ) ? $node : $this->wrapNamespace($node);
-    }
+//
```
theofidry: can be simplified

.TODO
```diff
@@ -0,0 +1,17 @@
+Right now whitelisted class works as follow:
```
theofidry: Should be transformed into several issues rather than this file

theofidry:
The e2e tests are not passing due to #152

PR #138, opened 2018-01-05
Create namespace for whitelisted classes

### LINKED ISSUE #135: Classes from global namespace incorrectly prefixed
Given the following class:

```
<?php

class AppKernel
{
}
```

which is whitelisted (hence should be prefixed), the prefixed code won't change i.e. the namespace with the prefix added is missing:

```php
<?php

namespace PhpScoperPefix; // <- this is missing

class AppKernel
{
}
```

moorscode:
Technical choices:

This is the simplest solution to the problem we could think of.

### Global namespace
~Decided to not put classes in a global namespace that are not whitelisted in a file, which contains a whitelisted class.~

~We have tried this, but it resulted in having to put every class in a `namespace { }` wrap, which fails a lot of tests.~

As PSR-0 declares:
`This means each class is in a file by itself, and is in a namespace of at least one level: a top-level vendor name.`

Updated the functionality to make sure any code that is not wrapped in a namespace will receive the global namespace. This prevents invalid PHP from being generated.

Paired on it together with @atimmer

Fixes #135


moorscode:
While testing this code in a project, we ran across the problem that defines are placed outside the namespace.

This is causing fatal errors.

Placing the define inside the namespace resolves the error.

moorscode:
Having to resolve this, I think we need to wrap -anything- in a namespace group as this is the method of prefixing used in this tool.

This would result in:
```php
namespace {
    define('SOME_DEFINE', 'value');
}

namespace a {
    class b {}
}

namespace {
    class c {}
}
```

specs/namespace/creation-for-whitelist.php
```diff
@@ -0,0 +1,85 @@
...
+    'meta' => [
+        'title' => 'Namespace declaration creation for whitelisted classes',
+        // Default values. If not specified will be the one used
+        'prefix' => 'Humbug',
+        'whitelist' => [
+        ],
+    ],
+
+    [
+        'spec' => <<<'SPEC'
```
theofidry: good use of the format:

```php
[
    'spec' => '',
    'payload' => '',
]
```

but I think it's a case simple enough where we can use the simplified format:

```php
'spec content' => 'payload content'
```

As you can find [here](https://github.com/humbug/php-scoper/blob/master/specs/misc.php#L34)

specs/namespace/creation-for-whitelist.php
```diff
@@ -0,0 +1,85 @@
...
+
+class AppKernel
+{
+}
+
+PHP
+    ],
+
+    [
+        'spec' => <<<'SPEC'
```
theofidry: I would add two more cases:

- One with multiple different namespaces like you can find [here](https://github.com/humbug/php-scoper/blob/master/specs/class/regular.php#L145) (it doesn't need to be the same it's just for the idea)
- One with two classes belonging to the global namespace: one which is a whitelisted one, so `AppKernel` which as a result will be prefixed, and another which is not like `Closure` which should not be prefixed

src/NodeVisitor/NamespaceStmtCreator.php
```diff
@@ -0,0 +1,93 @@
...
+namespace Humbug\PhpScoper\NodeVisitor;
+
+use Humbug\PhpScoper\NodeVisitor\Collection\NamespaceStmtCollection;
+use PhpParser\Node;
+use PhpParser\Node\Stmt\Class_;
+use PhpParser\Node\Stmt\Namespace_;
+use PhpParser\NodeVisitorAbstract;
+
+/**
+ * Creates a namespace for whitelisted classes.
```
theofidry: `[...] belonging to the global namespace`. It is an important part as the whitelisting feature is different depending of if the class belongs or not to the global namespace cf. [global-namespace-whitelisting](https://github.com/humbug/php-scoper#global-namespace-whitelisting) and [whitelist](https://github.com/humbug/php-scoper#whitelist)

src/NodeVisitor/NamespaceStmtCreator.php
```diff
@@ -0,0 +1,93 @@
...
+ * ```
+ * namespace Humbug;
+ *
+ * class AppKernel {}
+ * ```
+ */
+final class NamespaceStmtCreator extends NodeVisitorAbstract
+{
+    private $prefix;
+
```
theofidry: let's remove the blank lines between those properties

src/NodeVisitor/NamespaceStmtCreator.php
```diff
@@ -0,0 +1,93 @@
...
+    private $namespaceStatements;
+
+    private $globalWhitelister;
+
+    /**
+     * @param string                  $prefix
+     * @param NamespaceStmtCollection $namespaceStatements
+     * @param callable                $globalWhitelister
+     */
```
theofidry: the doc here is not useful as completely redundant with the signature, I know I may not have been consistent everywhere but I would remove it still :)

src/NodeVisitor/NamespaceStmtCreator.php
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function leaveNode(Node $node)
+    {
+        return ($node instanceof Class_)
+            ? $this->addNamespaceStmt($node)
+            : $node;
```
theofidry:
```php
return ($node instanceof Class_)
    ? $this->addNamespaceStmt($node)
    : $node
;
```

src/NodeVisitor/NamespaceStmtCreator.php
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    /**
+     * @param Class_ $node
```
theofidry: same here, useless phpdoc

specs/namespace/creation-for-whitelist.php
```diff
@@ -0,0 +1,85 @@
+<?php
```
theofidry: There is also a few other cases that you can find in `specs/class` that needs to be accounted for. For example it should also apply to:

- abstract classes
- interfaces
- ~~traits~~: they cannot be whitelisted; may require a test that it doesn't work there still
- ~~anonymous~~: they cannot be whitelisted; may require a test that it doesn't work there still

src/NodeVisitor/NamespaceStmtCreator.php
theofidry: Ah right, I need to figure out which parameter in the CS fixer it is to disable that, it really annoys me

moorscode:
Additional functionality:

I've used the `beforeTraverse` to determine if this "scope/file" has any whitelisted elements. If so, anything not white-listed needs to have a namespace.

If no namespace exists, the global namespace will be wrapped around it.

This resolves the problems described before.

moorscode:
(Sorry I'm not so good at squashing commits and pushing after everything is done ;))

theofidry:
>(Sorry I'm not so good at squashing commits and pushing after everything is done ;))

Don't worry about it I'll do a "Squash & Merge", only the CI may resent you a bit if you push them one by one but I think Travis & Scrutinizer got better at it

src/NodeVisitor/NamespaceStmtCreator.php
```diff
@@ -0,0 +1,125 @@
...
+        ) {
+            return new Namespace_(new Node\Name($this->prefix), [$node]);
+        }
+
+        // Anything else needs to be wrapped with global namespace.
+        return new Namespace_(null, [$node]);
+    }
+
+    /**
+     * @param Node[] $nodes List of nodes to find whitelisted nodes from.
```
theofidry: I think the signature is descriptive enough, no need for the `List [...]` comment. I would keep the phpdoc as it does add the type `array`, but I tend to avoid comments when the code is self-explanatory as the problem with comments/docs is that they easily get out of date with the code they are documenting. So keeping it to minimal is better IMO

src/NodeVisitor/NamespaceStmtCreator.php
```diff
@@ -0,0 +1,125 @@
...
+    {
+        $nodes = array_filter($nodes, function ($node) {
+            return null !== $node->name && ($this->globalWhitelister)($node->name);
+        });
+
+        return !empty($nodes);
+    }
+
+    /**
+     * @return bool
```
theofidry: no need for the phpdoc here

src/NodeVisitor/NamespaceStmtCreator.php
```diff
@@ -0,0 +1,125 @@
...
+     *
+     * @return bool
+     */
+    private function hasWhitelistedNode(array $nodes): bool
+    {
+        $nodes = array_filter($nodes, function ($node) {
+            return null !== $node->name && ($this->globalWhitelister)($node->name);
+        });
+
+        return !empty($nodes);
```
theofidry: as the returned type is an array I think `[] !== $nodes` or with a count is more expressive here

src/NodeVisitor/NamespaceStmtCreator.php
```diff
@@ -0,0 +1,125 @@
...
+        $this->globalWhitelister = $globalWhitelister;
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function beforeTraverse(array $nodes)
+    {
+        $classes = array_filter($nodes, [$this, 'isWhitelistableNode']);
+        if (empty($classes)) {
```
theofidry: as the returned type is an array I think [] !== $nodes or with a count is more expressive here

moorscode:
While writing specs, I came across another problem when a global namespace is already defined around a class.
It does not seem to be in the scope of this feature, so I decided not to add it in this PR:

```
'Traits in different namespace.' => <<<'PHP'
<?php

namespace Foo {
    trait SomeTrait{}
}

namespace {
    class AppKernel{}
}

----
<?php

namespace Humbug\Foo{
    trait SomeTrait
    {
    }
}
namespace Humbug {
    class AppKernel
    {
    }
}

PHP
```

As the namespace prefixer should pick this up and add the namespace for the whitelisted class in the global namespace.

src/NodeVisitor/NamespaceStmtCreator.php
moorscode: Do you prefer to have an empty docblock or just none at all?

theofidry:
You mean right now we would have:

```php
<?php

namespace Foo {
    trait SomeTrait{}
}

namespace {
    class AppKernel{}
}

----
<?php

namespace Humbug\Foo{
    trait SomeTrait
    {
    }
}
namespace Humbug {
    class AppKernel
    {
    }
}
```

instead of having canonical namespaces, i.e. only one `Humbug` namespace declaration? If that's the case IMO there is nothing to do here it works fine https://3v4l.org/DCjXW

src/NodeVisitor/NamespaceStmtCreator.php
moorscode: In the projects we do at Yoast, we have very strict documentation rules.
We don't find ourselves having a lot of outdated documentation entries in general.

The fact that we are in a WordPress environment, where backwards compatibility is a major thing might have something to do with it, but I don't think that is the only factor in it.

Writing the documentation mostly helps me re-evaluate what I wrote functionally and if the naming of the variables and logic are actually the best solutions.

My two cents on the subject.

I'm happy to oblige to the standards and approach you set for your project.

theofidry: I think `@param Node[] $nodes` is enough here

If it's too "redundant" with the code, I really prefer to avoid the comments. For public interfaces, non trivial things or for example the doc to explain the purpose of the class I tend to be in favour comments :)

moorscode:
Hope this output helps to describe it a bit better.
The term `whitelist` gets a bit confusing as it whitelists from receiving the prefix..

I would expect the `AppKernel` (whitelisted node name) to be prefixed.
But as it already has a namespace, which is global, it is ignored in the Namespace Statement Prefixer: https://github.com/humbug/php-scoper/blob/master/src/NodeVisitor/NamespaceStmtPrefixer.php#L79
```
================================================================================
INPUT
================================================================================
<?php

namespace Foo {
    trait SomeTrait{}
}

namespace {
    class AppKernel{}
}

namespace {
    class Bla{}
}


================================================================================
EXPECTED
================================================================================
<?php

namespace Humbug\Foo {
    trait SomeTrait
    {
    }
}
namespace Humbug {
    class AppKernel
    {
    }
}
namespace {
    class Bla
    {
    }
}


================================================================================
ACTUAL
================================================================================
<?php

namespace Humbug\Foo {
    trait SomeTrait
    {
    }
}
namespace {
    class AppKernel
    {
    }
}
namespace {
    class Bla
    {
    }
}
```

theofidry:
We're clear on the expected input/expected output; But I wonder if the work you're doing shouldn't be in the `NamespaceStmtPrefixer` as otherwise, this one is missing the necessary context to know if the namespace should be prefixed or not

moorscode:
> We're clear on the expected input/expected output; But I wonder if the work you're doing shouldn't be in the NamespaceStmtPrefixer as otherwise, this one is missing the necessary context to know if the namespace should be prefixed or not

That's a good point, I initially started a new visitor because prefixing and creating seemed like two concepts.
As the global namespace doesn't have to be defined but is always present, prefixing it and making it explicit fits in the `prefix` concept when looking at it that way.

moorscode:
Updated the code and added the desired StyleCI disables to remove the checks on property separation and const separation.

moorscode:
Also put it in a separate PR: https://github.com/humbug/php-scoper/pull/139

src/NodeVisitor/NamespaceStmtPrefixer.php
```diff
@@ -38,15 +40,30 @@
 {
     private $prefix;
     private $namespaceStatements;
+    private $hasWhitelistedNode;
+    private $globalWhitelister;
 
     /**
      * @param string                  $prefix
      * @param NamespaceStmtCollection $namespaceStatements
+     * @param callable                $globalWhitelister
```
theofidry: let's just remove the phpdoc here

specs/namespace/creation-for-whitelist.php
```diff
@@ -0,0 +1,300 @@
...
+    }
+}
+
+PHP
+    ,
+
+    'Defines should be wrapped in namespace alongside whitelisted class.' => <<<'PHP'
+<?php
+
+define( "MY_DEFINE", "value" );
```
theofidry: small extra space here :)

theofidry:
Many thanks @moorscode, that was a critical bug :)

PR #139, opened 2018-01-06
StyleCI: new line checks

moorscode:
Added the desired StyleCI disables to remove the checks on property separation and const separation.

This will prevent the check from failing.


theofidry:
Cheers :)

PR #140, opened 2018-01-10
Restore required PHP version to 7.0

moorscode:
This was introduced in https://github.com/humbug/php-scoper/commit/d2fb4d4ecdbc2b5c2db8e7fceffbe792a0a5cd67#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780R21

But the dependency that cause this bump has been removed afterwards.
[`myclabs/deep-copy`](https://github.com/humbug/php-scoper/commit/d2fb4d4ecdbc2b5c2db8e7fceffbe792a0a5cd67#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780R22)

Removed in: https://github.com/humbug/php-scoper/commit/0b4843bef010e7912ca31558627c971e614c8531#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780L22

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,9 +8,9 @@ cache:
 
 matrix:
   include:
-    - php: '7.1'
+    - php: '7.0'
       env: COVERAGE='true'
-    - php: '7.1'
+    - php: '7.0'
     - php: nightly
   allow_failures:
     - php: nightly
@@ -48,7 +48,7 @@ deploy:
   on:
     tags: true
     repo: humbug/php-scoper
-    php:  '7.1'
+    php:  '7.0'
     condition: "$COVERAGE"
 
 notifications:
--- a/Makefile
+++ b/Makefile
@@ -58,7 +58,7 @@ tc: vendor/bin/phpunit
 	phpdbg -qrr -d zend.enable_gc=0 $(PHPUNIT) --coverage-html=dist/coverage --coverage-text
 
 e2e:		## Run end-to-end tests
-e2e: e2e_004 e2e_005 e2e_011 e2e_013
+e2e: e2e_004 e2e_005 e2e_011 e2e_013 e2e_014 e2e_015
 
 e2e_004:	## Run end-to-end tests for the fixture set 004: source code case
 e2e_004: bin/php-scoper.phar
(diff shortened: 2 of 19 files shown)


theofidry:
Should we though? PHP 7.0 is in security fixes until the end of '19 so it won't be around for much more. Also box will be pushed to 7.1 as well if that's not already the case so I'm not sure it's worth trying to keep 7.0

moorscode:
Forcing people to use a later version of PHP does not sound like a very friendly thing to do.
There are many people/environments that don't have direct control over the version of PHP they can run on.

You never limit the PHP version if you lower it to that lowest version.

From a more selfish perspective I'd like to share the situation we want to use this package in.

We build a WordPress plugin and we run on 7+ million sites.
WordPress is compatible with PHP 5.2.4+.
We are very aware of all the different situations people are running the sites on and developing their plugins, themes or other integrations on.

As our plugin is Open Source, we welcome and encourage anybody to help develop it.
https://github.com/Yoast/wordpress-seo/

Developer happiness is something we value very much, and as such want to make sure the lowest possible version is the restriction we lay upon any contributors or developers.

If we can provide a minimal PHP version of 7.0 instead of 7.1 (or higher) that would mean a great deal to us. We currently have a minimal requirement of PHP 5.3.2, as this is the minimal version required to actually run composer.

WordPress is also the reason why we need to be able to prefix dependencies, as any other plugin or theme could potentially use the same dependency.

This has caused us problems in the past, with conflicting versions of packages being installed. Depending on the loading order of the plugin, a certain version would be loaded and used by all other plugins requiring this functionality.

theofidry:
Maybe I can downgrade to 7.0 for this library it's not that big of a deal, but as you say it may not be enough as Wordpress is compatible with 5.2.4+ and there is a question of where you put the limit.

IMO a more viable alternative is to ship PHP-Scoper in a docker or VM as it requires 7.1 but is perfectly capable of scoping PHP 5.2 code. Wouldn't that suit your case better?

moorscode:
Let me explain the different situations we need to work with a little more.

WordPress.org holds the build version of our plugin and provides a zip-file to install it on a specific WordPress installation.
This needs to have the dependencies prefixed, as multiple versions of the plugin or other plugins with the same dependencies can be installed simultaneously.
_There is no version/dependency management in WordPress itself._

As the code is Open Source, developers should be able to checkout our repository and work on developing it.
We don't want to commit the prefixed dependencies to the repository, so we need to provide a way to build them.
This is being done with a PHP script to determine if the `install` command is ran with `no-dev` or with dev. When we are running in `dev` mode, the `prefix-dependencies` composer script is triggered, which runs `php-scoper` on the specific dependencies.

We have implemented dynamic class-aliasing for situations where the repository is not checked out as a root package, as we then have a version management system to make sure we don't collide with other packages.

Thus we want to make sure we provide the biggest range of PHP versions locally installed, to make sure as many developers as possible are able to check out the repository and run `composer install`.

This is the main reason why we strongly prefer to have `7.0` over `7.1`, especially when the code does not technically require `7.1` to run.

We find it hard to accept that this is the reality we live in as we also like to be on the latest, safest and most feature-complete version of PHP, but we just can't demand that and exclude the people that cannot (easily) meet these requirements.

atimmer:
Hey @theofidry, PHP 5.2 is our burden. Be happy it isn't yours 😉. No sane person would want to support this, but it is reality in the WordPress community. But I don't think anyone should be on the other extreme either. It is worth for libraries to be compatible with a lot of PHP versions, given those are supported by PHP, because this improves the accessibility of the library. And the burden is not too high.

Given this PR has no cost for you, I don't see a reason not to merge it, unless I am missing something. The end of '19 is still almost two years away.

theofidry:
Fair enough, could you update Travis accordingly? I wouldn't want to break something for 7.0 by accident

moorscode:
It seems that I did not run the tests while on PHP 7.0 locally.
There are too many PHP 7.1 constructs being used in the code now to easily go back to PHP 7.0.

The things that I ran into:
- private class constants
- nullable return value definitions (`?string`)

theofidry:
Another issue I noticed now is https://github.com/Roave/BetterReflection requires 7.1 which will be a requirement for #117 :/

I think the docker way would be the smoothest for you, this will allow you to make it work from 7.0 down to 5.2 without any hurdle.

PR #141, opened 2018-01-11
Support longer PHP version strings

firstred:
It seems that PHP version strings do not necessarily have to follow SemVer standards. Running this package on Ubuntu, for example, causes the following fatal error:
```
PHP Fatal error:  Uncaught UnexpectedValueException: Invalid version string "7.1.12-3" 
```
The full version string is: `7.1.12-3+ubuntu16.04.1+deb.sury.org+1`

Therefore, replacing composer's SemVer comparison with PHP's native `version_compare` might be a better idea to support these platforms, too.

### DIFF
--- a/src/Autoload/Requirements.php
+++ b/src/Autoload/Requirements.php
@@ -42,7 +42,7 @@ final class Requirements extends RequirementCollection
         $requiredPhpVersion = $composerConfig['php'];
 
         $this->addRequirement(
-            Semver::satisfies(phpversion(), $requiredPhpVersion),
+            version_compare(phpversion(), $requiredPhpVersion, '>='),
             sprintf(
                 'PHP version must satisfy "%s" ("%s" installed)',
                 $requiredPhpVersion,


theofidry:
Thanks for the bug report! A couple of questions:

- How does Composer handle this? I used the `SemverPackage` because I thought it would cover that kind of edge case
- I think it's the only place where `Semver` is used so we can get rid of that dependency if that's the case
- I'm fine with the patch but a bit worried by the lack of tests; be it an e2e or unit test, I think we should have something :)

firstred:
Thanks for your quick reply.

> How does Composer handle this? I used the SemverPackage because I thought it would cover that kind of edge case

`SemverPackage` unfortunately does not handle such an edge case. It handles perfect SemVer strings. Since `version_compare` has been introduced to " Compares two "PHP-standardized" version number strings" I figured that would be a lot better and in my case it is.

> I think it's the only place where Semver is used so we can get rid of that dependency if that's the case

Always nice and I think it is possible to get rid of the dependency.

> I'm fine with the patch but a bit worried by the lack of tests; be it an e2e or unit test, I think we should have something :)

Not sure what you mean by that? Do you think we should test built-in PHP functionality or keep it aligned with the way composer handles it?
I am not very familiar with the way composer compares PHP versions, but it looks like it does some additional cleaning before applying the SemVer comparison, but that is because it has to handle SemVer compatibility strings as well.
In this case we only need to clean it up and compare in a SemVer way, which is exactly what we can achieve by using `version_compare`.

theofidry:
>>I'm fine with the patch but a bit worried by the lack of tests; be it an e2e or unit test, I think we should have something :)

>Not sure what you mean by that?

I mean that right now everything is green so a quick refacto can easily introduce a regression there so we need a test of some kind, for example adding a build on Travis with a specific distrib for which the issue occurs or a unit test

I didn't dig into this so I'm not sure on how to do either of them :)

TomasVotruba:
I got same issue. Why is `Composer\Semver` even needed here?

theofidry:
@TomasVotruba from my first comment, it's no longer needed if we apply this patch so it should be removed there as well IMO

TomasVotruba:
This was closed by #154, but not automatically, cause Github needs the keyword before the id. Probably :)

theofidry:
Indeed thanks :)

Issue #127, opened 2017-11-09
PSR-0 dependencies are not correctly prefixed


borNfreee:
| Question    | Answer
| ------------| ---------------
| PHP-Scoper version | master https://github.com/humbug/php-scoper/commit/1ca326df196e2afb85f0b79895fb7026fc85a3dd
| PHP version | 7.1.4
| Composer version | 1.5.2 2017-09-11 16:59:25
| Github Repo | https://github.com/borNfreee/php-scoper-bug

I'm trying to prefix [Infection](https://github.com/infection/infection/pull/67) application but faced an issue with PSR-0 dependency which is not correctly resolved.

I've created a github repo with reproducible issue - just a very small application with only 1 dependency - Pimple container which uses PSR-0 in its `composer.json`.

How to reproduce:

```bash
git clone https://github.com/borNfreee/php-scoper-bug
cd php-scoper-bug

composer install

# check it works
php src/main.php // prints "Hello"

# run PHP-Scoper
php-scoper add-prefix --output-dir=build-prefixed --force --prefix=Mutant

composer dump-autoload --working-dir=build-prefixed/ --classmap-authoritative --no-dev

# Then try to run prefixed application:
cd build-prefixed

php src/main.php
```

This throws an error:

```php
PHP Fatal error:  Uncaught Error: Class 'Mutant\Pimple\Container' not found in php-scoper-bug/build-prefixed/src/Application.php:10
Stack trace: ...
```

Please note that composer create `autoload_static.php` with the following content for PSR-0 dependencies:

```
public static $prefixesPsr0 = array (
        'P' => 
        array (
            'Pimple' =>  // <------- this is not prefixed
            array (
                0 => __DIR__ . '/..' . '/pimple/pimple/src',
            ),
        ),
    );
```

while PSR-4 dependencies are prefixed:

```bash
public static $prefixDirsPsr4 = array (
        'Mutant\\ScoperBug\\' => 
        array (
            0 => __DIR__ . '/../..' . '/src',
        ),
        'Mutant\\Psr\\Container\\' => 
        array (
            0 => __DIR__ . '/..' . '/psr/container/src',
        ),
    );
```

Did I miss something?

Hope this helps!

theofidry:
Indeed, looks like I completely forgot about PSR-0 [there](https://github.com/humbug/php-scoper/blob/master/src/Scoper/Composer/AutoloadPrefixer.php#L43)

rayrutjes:
[...]

PR #142, opened 2018-01-16
Add psr-0 support

BackEndTea:
This PR:

* [x] Adds support for PSR 0

Related to #127

### DIFF
--- a/src/Scoper/Composer/AutoloadPrefixer.php
+++ b/src/Scoper/Composer/AutoloadPrefixer.php
@@ -40,6 +40,13 @@ final class AutoloadPrefixer
 
     private static function prefixAutoloads(array $autoload, string $prefix): array
     {
+        if (isset($autoload['psr-0'])) {
+            foreach(self::changePSRZeroToPSRFour($autoload['psr-0']) as $key => $value) {
+                $autoload['psr-4'][$key] = $value;
+            }
+            unset($autoload['psr-0']);
+        }
+
         if (isset($autoload['psr-4'])) {
             $autoload['psr-4'] = self::prefixAutoload($autoload['psr-4'], $prefix);
         }
@@ -53,6 +60,31 @@ final class AutoloadPrefixer
 
         foreach ($autoload as $namespace => $paths) {
             $loader[sprintf('%s\\%s', $prefix, $namespace)] = $paths;
+        }
+
+        return $loader;
+    }
+
+    private static function changePSRZeroToPSRFour(array $psrZero): array
+    {
+        $loader =[];
+
+        foreach($psrZero as $namespace => $path) {
+
+            //Append backslashes, if needed, since psr-0 does not require this
+            if (substr($namespace, -1) !== '\\') {
+                $namespace .= '\\';
+            }
+
+            //Append a slash to the path if it does not have it
+            if (substr($path, -1) !== '/') {
+                $path .= '/';
+            }
+
+            $namespaceForPsr = str_replace('\\', '/',$namespace);
+            $path .= $namespaceForPsr . '/';
+
+            $loader[$namespace] = $path;
         }
 
         return $loader;
--- a/tests/Scoper/Composer/JsonFileScoperTest.php
+++ b/tests/Scoper/Composer/JsonFileScoperTest.php
@@ -171,4 +171,76 @@ JSON
 JSON
         ];
     }
+
+    /**
+     * @dataProvider providePSRZeroComposerFiles
+     */
+    public function test_it_prefixes_psr0_autoloaders(string $fileContent, string $expected)
+    {
+        touch($filePath = escape_path($this->tmp.'/composer.json'));
+        file_put_contents($filePath, $fileContent);
+
+        $scoper = new JsonFileScoper(new FakeScoper());
+
+        $prefix = 'Foo';
+        $patchers = [create_fake_patcher()];
+        $whitelist = ['Foo'];
+        $whitelister = create_fake_whitelister();
+
+        $actual = $scoper->scope($filePath, $prefix, $patchers, $whitelist, $whitelister);
+
+        $this->assertSame($expected, $actual);
+    }
+
+    public function providePSRZeroComposerFiles()
+    {
+        yield [
+            <<<'JSON'
+{
+    "bin": ["bin/php-scoper"],
+    "autoload": {
+        "psr-0": {
+            "Humbug\\PhpScoper\\": "src/"
+        },
+        "files": [
+            "src/functions.php"
+        ]
+    },
+    "autoload-dev": {
+        "psr-0": {
+            "Humbug\\PhpScoper\\": "tests/"
+        },
+        "files": [
+            "tests/functions.php"
+        ]
+    }
+}
+
+JSON
+            ,
+            <<<'JSON'
+{
+    "bin": [
+        "bin\/php-scoper"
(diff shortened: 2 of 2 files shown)


theofidry:
Thanks very much for looking into this one @BackEndTea :)

I think however I don't trust those tests enough and we should go a step further an have an integration test for them. You have several examples of end-to-end tests for the PSR-4 ones which you can find in `fixtures/setXX` and the `Makefile`. My experience is that real life cases may differ a bit from those unit tests so an end-to-end tests is necessary.

Another case which is not considered here either is when the transformed into PSR-4 PSR-0 entry actually already exists: how should we handle this? If there is a reasonable and easy way we can handle it, otherwise I think it would be ok to throw off the towel here and simply bail out with a user-friendly message to the user

BackEndTea:
> Another case which is not considered here either is when the transformed into PSR-4 PSR-0 entry actually already exists: how should we handle this?

Could you give an example of this? And i can't think of a good way to handle this, so any ideas are welcome

theofidry:
Sure, consider if you have:

```json
{
  "require": {
    "psr-0" {
      "Acme": "src/"
    },
    "psr-4" {
      "Acme\\": "lib/"
    }
  }
}
```

Actually there is a graceful way to handle this:

```json
{
  "require": {
    "psr-4" {
      "Acme\\": ["src/", "lib/"]
    }
  }
}
```

BackEndTea:
@theofidry I updated this pr to handle the merging of psr-0 and 4, even if one of them is an array.

I have looked at the fixtures folder and the makefile, but i can't really make sense of what is happening there. So any help would be appreciated.

BackEndTea:
One case that is currently not supported is how psr-0 is used to load `Pimple` in version 1.1.1
https://github.com/silexphp/Pimple/tree/1.1
The composer is as follows:
```json
"autoload": {
    ""psr-0": { "Pimple": "lib/" }
}
```
And the file structure is as follows:
root
* 📁lib
  *  📋Pimple.php

So the psr-0 refers to a file rather than a directory.

Makefile
```diff
@@ -95,6 +95,24 @@ e2e_013: bin/php-scoper.phar
 	$(PHPSCOPER) init --working-dir=build/set013 --no-interaction
 	diff src/scoper.inc.php.tpl build/set013/scoper.inc.php
 
+e2e_014:	## Run end-to-end tests for the fixture set 014: source code case with psr-0
```
theofidry: can you double check that the indent is correct for the lines with `##` when doing `make`?

Makefile
```diff
@@ -140,6 +160,9 @@ vendor-bin/box/composer.lock: composer.lock
 fixtures/set005/composer.lock: fixtures/set005/composer.json
 	@echo fixtures/set005/composer.lock is not up to date.
 
+fixtures/set015/composer.lock: fixtures/set015/composer.json
```
theofidry: I would put it after `fixtures/set011`

fixtures/set014/composer.json
```diff
@@ -0,0 +1,10 @@
+{
+    "bin": [
+        "bin/greet"
+    ],
+    "autoload": {
+        "psr-0": {
+            "Set014\\": "src/"
+        }
+    }
+}
```
theofidry: missing return line here

fixtures/set015/composer.json
```diff
@@ -0,0 +1,13 @@
...
+    ],
+    "autoload": {
+        "psr-0": {
+            "Set015\\": "src/"
+        }
+    },
+    "require": {
+        "pimple/pimple": "3.2.3"
+    }
+}
```
theofidry: missing return line here

src/Scoper/Composer/AutoloadPrefixer.php
```diff
@@ -57,4 +64,90 @@ private static function prefixAutoload(array $autoload, string $prefix): array
 
         return $loader;
     }
+
+    private static function mergePSRZeroAndFour(array $psrZero, array $psrFour): array
```
theofidry: let's change it to `mergePSR0And4`

src/Scoper/Composer/AutoloadPrefixer.php
```diff
@@ -57,4 +64,90 @@ private static function prefixAutoload(array $autoload, string $prefix): array
...
+    {
+        foreach ($psrZero as $namespace => $path) {
```
theofidry: `$psr0`

src/Scoper/Composer/AutoloadPrefixer.php
```diff
@@ -57,4 +64,90 @@ private static function prefixAutoload(array $autoload, string $prefix): array
...
+
+                continue;
+            }
+            $psrFour[$namespace] = self::mergeNamespaces($namespace, $path, $psrFour);
+        }
+
+        return $psrFour;
+    }
+
+    private static function updatePSRZeroPath($path, $namespace)
```
theofidry: `updatePSR0Path`

src/Scoper/Composer/AutoloadPrefixer.php
```diff
@@ -57,4 +64,90 @@ private static function prefixAutoload(array $autoload, string $prefix): array
...
+    {
+        foreach ($psrZero as $namespace => $path) {
+            //Append backslashes, if needed, since psr-0 does not require this
+            if ('\\' !== substr($namespace, -1)) {
+                $namespace .= '\\';
+            }
+
+            $path = self::updatePSRZeroPath($path, $namespace);
+
+            if (!isset($psrFour[$namespace])) {
```
theofidry: `$psr4`

src/Scoper/Composer/AutoloadPrefixer.php
```diff
@@ -57,4 +64,90 @@ private static function prefixAutoload(array $autoload, string $prefix): array
...
+     * string     |
+     * or simply the namepace not existing as a psr-4 entry.
+     *
+     * @param string$psrZeroNamespace
+     * @param string|array $psrZeroPath
+     * @param string|array $psrFour
+     *
+     * @return string|array
+     */
+    private static function mergeNamespaces(string $psrZeroNamespace, $psrZeroPath, $psrFour)
```
theofidry: `mergeNamespaces(string $psr0Namespace, $psr0Path, $psr4)`

src/Scoper/Composer/AutoloadPrefixer.php
```diff
@@ -57,4 +64,90 @@ private static function prefixAutoload(array $autoload, string $prefix): array
...
+    }
+
+    /**
+     * Deals with the 4 possible scenarios:
+     *       PSR0 | PSR4
+     * array      |
+     * string     |
+     * or simply the namepace not existing as a psr-4 entry.
+     *
+     * @param string$psrZeroNamespace
```
theofidry: missing a space here

tests/Scoper/Composer/JsonFileScoperTest.php
```diff
@@ -168,6 +168,226 @@ public function provideComposerFiles()
         ]
     }
 }
+JSON
+        ];
+    }
+
+    /**
+     * @dataProvider providePSRZeroComposerFiles
```
theofidry: `providePSR0ComposerFiles`

theofidry:
Awesome, thanks very much for this 🏅

PR #144, opened 2018-01-29
Rework config and command

theofidry:
Reworked the configuration to make it easier to re-use: the command now
provides the globalNamespaceWhitelister ready for use as well as the
list of files with their contents.

As the handler has a lot of IO-bound dependencies, it is actually very
hard to re-use so it makes more sense to integrate it back to the
command.

This paves the way for https://github.com/humbug/box/pull/31.


tests/Handler/HandleAddPrefixTest.php
```diff
@@ -12,393 +12,405 @@
...
-                '/dir2/dir3/unknown',
-                '/dir2/fileB.php',
-                '/dir2/fileC.php',
-                '/dir2/unknown',
-                '/file2.php',
-            ],
-        ];
-    }
-}
+//declare(strict_types=1);
```
theofidry: this is left to do

Issue #152, opened 2018-02-15
Handle internal constants & functions


theofidry:
Constant and functions should be prefixed unless they are internal.

theofidry:
Closed by #155 and #117

PR #155, opened 2018-02-16
Fix prefixing of functions

### LINKED ISSUE #145: Not working with imported built-in functions
```php
use function getcwd();
```

is going to be prefixed:

```php
use function PhpScoperXXX\getcwd();
```

which is obviously wrong

theofidry:
Now prefix functions whenever possible. A function is not prefixed when is an internal function or when the function fully qualified name could not be resolved.

Related to #152.
Closes #145.

The failure on the e2e tests is unrelated.

### DIFF
--- a/src/Reflector.php
+++ b/src/Reflector.php
@@ -14,16 +14,21 @@ declare(strict_types=1);
 
 namespace Humbug\PhpScoper;
 
+use ReflectionException;
+use ReflectionFunction;
 use Roave\BetterReflection\Reflector\ClassReflector;
 use Roave\BetterReflection\Reflector\Exception\IdentifierNotFound;
+use Roave\BetterReflection\Reflector\FunctionReflector;
 
 final class Reflector
 {
     private $classReflector;
+    private $functionReflector;
 
-    public function __construct(ClassReflector $classReflector)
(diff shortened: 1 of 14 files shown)


src/Reflector.php
```diff
@@ -34,4 +39,16 @@ public function isClassInternal(string $name): bool
             return false;
         }
     }
+
+    public function isFunctionInternal(string $name): bool
+    {
+        try {
+            return (new ReflectionFunction($name))->isInternal();
+
+            // TODO: leverage Better Reflection instead
```
theofidry: @Ocramius could you double check that? I may have misunderstood how it works but from what I see the identifier is correctly built but `PhpInternalSourceLocator` just tries to get the class from that identifier (which fails here as it's a function). 

See [source](https://github.com/Roave/BetterReflection/blob/master/src/SourceLocator/Type/PhpInternalSourceLocator.php#L42)

Ocramius: Aye, that looks like a bug in BR

theofidry:
I'll do the switch to BR when it will be fixed there then :) I reported the issue in https://github.com/Roave/BetterReflection/issues/407

PR #172, opened 2018-02-22
Add Infection

theofidry:
Add Infection to the CI. However as it would be too boring to stop at that, we also run the scoped version.


theofidry:
The failure seems unrelated so I'm merging this :)

PR #176, opened 2018-02-27
Fix Infection tests

theofidry:
Infection was "passing" only because the process was failing hard which indeeds counts as killing a mutant, but in this specific case this was not for the right reasons.

I investigated and managed to apply the right patches to make it work. The e2e tests also no runs the not prefixed Infection and compares the result with the prefixed one as they should be identical.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ help:
 
 clean:		## Clean all created artifacts
 clean:
-	git clean --exclude=.idea/ -fdx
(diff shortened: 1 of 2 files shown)


theofidry:
@borNfreee with it it's ready to go

borNfreee:
Awesome 👍 

We will definitely use your work soon after I merge https://github.com/infection/infection/pull/67
I'm going to release 0.8.0 and the next thing for me will be integrating php scoper

PR #213, opened 2018-06-05
[WIP] Allow to whitelist a namespace

### LINKED ISSUE #192: Whitelisting
Hi!

At the PHPUnit Code Sprint here in Hamburg yesterday and today, @kambo-1st worked on https://github.com/sebastianbergmann/phpunit/issues/2015. Based on his findings, we came to the conclusion that (at least) the {classes|interfaces|traits} of `phpunit/phpunit` and `phpunit/phpunit-mock-objects` need to be whitelisted when building a prefixed PHAR. By whitelisting we mean that the names of these units of code are not prefixed. The code that declares these units of code needs to be processed, though, as it need to be updated to used the prefixed names of the other dependencies.

A problem that we ran into is the fact that php-scoper requires a list of the classes to be whitelisted. Is there any way that whitelisting could work based on namespace names or (FQCN) class name prefixes?

theofidry:
Closes #192

Allow to whitelist a complete namespace with the `*` notation, e.g. `*` (for the global namespace) or `Acme\*`

**[BC break]:** the API signature has changed to convert the `array` `$whitelist` into the `Whitelist` object


theofidry:
Still missing a unit test for the whitelist and ready to go

theofidry:
The CI failure is expected: there is a BC break breaking Box which needs to be updated

PR #244, opened 2018-07-04
Fix the CI

theofidry:
- Disable xdebug
- Fix PHP-Scoper version to allow the newer versions of Box to be installed
- Fix Scrutinizer
- Bump the dependencies
- Ensure the dependencies are compatible for PHP 7.1.0

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -1,0 +1,4 @@
+build:
+    environment:
+        variables:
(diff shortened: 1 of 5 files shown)


theofidry:
Scrutinizer doesn't seem to check the config file, I'm requesting an access to the repo on Scrutinizer to fix this, moving forward meanwhile

PR #245, opened 2018-07-04
Fixed docs about prefixing

Wirone:
Use correct representation of empty string, removed duplicated word.


theofidry:
Many thanks! :)

PR #271, opened 2018-10-26
Setup the fixtures for the Symfony support

### LINKED ISSUE #182: Add frameworks integration
For scoping a Symfony application for example, right now extra manual steps are required to scope the service files. This kind of thing could be handled by PHP-Scoper

- [ ] Symfony
- [x] Laravel

theofidry:
Closes #182

- [x] Yaml support
  - [x] Class param
  - [x] Class as service ID
  - [x] Class as alias
  - [x] Class as alias shortcut
  - [x] Class as argument
  - [x] Class as tag element
  - [x] Whitelisted class
  - [x] Class belonging to a whitelisted namespace
- [x] XML support
  - [x] Class param
  - [x] Class as service ID
  - [x] Class as alias
  - [x] Class as alias shortcut
  - [x] Class as argument
  - [x] Class as tag element
  - [x] Whitelisted class
  - [x] Class belonging to a whitelisted namespace
- [ ] PHP support
  - [ ] Class param
  - [ ] Class as service ID
  - [ ] Class as alias
  - [ ] Class as alias shortcut
  - [ ] Class as argument
  - [ ] Class as tag element
  - [ ] Whitelisted class
  - [ ] Class belonging to a whitelisted namespace
- [ ] Neon support?


TomasVotruba:
This would be great :+1:  :)

Indeed, mainly for `services.yml`

At the moment I have to do it with regular expression like this: https://github.com/rectorphp/prefixer/blob/a10a2a005673e72b4beec4f4611b3b5651e2ca92/src/Worker/PrefixConfigsWorker.php#L29-L33

I still need it for `services.neon` files, because they're in `/vendor` as well.

theofidry:
Let's go with the YAML and XML support first. The PHP one should already somewhat be there and for the Neon one I'm not sure but I'll be happy to add it afterwards.

TomasVotruba:
Agreed, I'd be to happy to test it ASAP.

As for Symfony WTFs, I've found that container dump needs to be prefixed extra, since it's string: https://github.com/rectorphp/prefixer/blob/master/src/PrefixFixer.php#L50-L61

TomasVotruba:
I'm giving this try after year and still end up with same error.

Here is the code we use: https://github.com/rectorphp/rector/pull/2373

Autoconfigure configs are not prefixed:

```yaml
services:
    _defaults:
        public: true
        autowire: true

    Rector\:
        resource: '../src'
```


They should be, right?

```
        resource: 'phar://../src'
```

Because when I run:

```bash
../tmp/rector.phar 
```

I get this error:


```bash
[ERROR] The                                                                                                            
         "phar:///var/www/rector/tmp/rector.phar/src/HttpKernel/../../config/../packages/NetteToSymfony/config/../src/Route/RouteInfo.php" directory does not exist in phar:///var/www/rector/tmp/rector.phar/src/HttpKernel/../../config/../packages/NetteToSymfony/config/config.yaml (which is being imported from "phar:///var/www/rector/tmp/rector.phar/src/HttpKernel/../../config/config.yaml").  
```

TomasVotruba:
```bash
Box version 3.8.3@43f13de 2019-11-03 17:18:32 UTC
```

TomasVotruba:
After short exploration... it seems it's another Symfony bug.
The error is caused here: https://github.com/symfony/symfony/blob/c62bb82e27974ef4e389da523f0de451b6632266/src/Symfony/Component/Finder/Finder.php#L589

This was similar fix, might be related: https://github.com/symfony/symfony/commit/b61c8fcb77282f480941026fabcbc112b8a7f50e

theofidry:
@TomasVotruba it's been a while so I may remember wrong, but it looks weird to me. The bug you report makes sense to me in the context of the bundled application, i.e. the PHAR of Rector, _compiles_ its containers. This should not happen, see https://github.com/humbug/box/blob/master/doc/symfony.md#symfony-support for a more detailed explanation as to why

TomasVotruba:
It's a Symfony bug: https://github.com/symfony/symfony/blob/c62bb82e27974ef4e389da523f0de451b6632266/src/Symfony/Component/Finder/Finder.php#L589

Not sure where exactly it stars thought

theofidry:
It might be but I'm point is the reason why you're hitting that bug in the first place might be the real issue here, at least for your case.

If the Symfony compiler compiles within the PHAR, it can never work because the PHAR is supposed to be read only

TomasVotruba:
Oh. So what would be the solution?
I didn't get that from the link

TomasVotruba:
There are no assets, just services

theofidry:
>There are no assets, just services

I'm not really following how that is related. I'm suggesting - I don't know if that's the real issue, that when you package Rector in the PHAR and then use it, because it has not been packaged "properly" (as a Symfony app requires a bit more tweaking up-front), it attemps to compile the container which will 200% fail and will never ever work by design. A PHAR is a ready-only environment so compiling the container which dumps various files in its cache directory won't work out.

The suggestions in the article are to prevent that:

- make sure that your env variables are dumped if any
- make sure your assets are dumped if any
- make sure your cache is warmed up (which will ensure the compiler is compiled
- make sure that all the above happens in prod mode with debug set to `false`
- make sure that your app running in the PHAR _remains_ in prod mode with debug set to `false`

TomasVotruba:
Rector doesn't have any env variables. That's why I don't get it.
Also it's container is dumped in normal way to temp cache. Like any other cache, that changes based on the configuration so it can't be persisted.

There were 2 issues with Symfony config resource/exclude bugs in PHAR. Nothing more yet. I've found saint workarounds that compensate bugs.

Finally the container is build and app run, yay:

good_times


Now more testing is coming :)

PR #300, opened 2019-02-19
Fixed duplicate key found in scoper.inc.php.tpl

poisa:
The key `whitelist-global-constants` was found twice. I replaced one of them to be `whitelist-global-classes`.


theofidry:
Cheers

PR #310, opened 2019-04-17
 "vendor" directories matching pattern adjustment

kmgalanakis:
Update the pattern that matches "vendor" directories to work on Windows.


theofidry:
Thanks!

PR #318, opened 2019-05-06
Leverage BetterReflection and PHPStorm stubs

theofidry:
Waiting on https://github.com/Roave/BetterReflection/pull/471


tests/ReflectorTest.php
```diff
@@ -101,62 +108,51 @@ class DateTime {}
...
-            <<<'PHP'
-<?php
-
-function class_exists {}
-PHP
-            ,
-            'class_exists',
-            true,
-        ];
+//        yield [
```
theofidry: to be reverted

tests/ReflectorTest.php
```diff
@@ -210,11 +212,34 @@ private function createReflector(string $code): Reflector
 
         $sourceLocator = new AggregateSourceLocator([
             new StringSourceLocator($code, $astLocator),
-            new PhpInternalSourceLocator($astLocator),
+            new PhpInternalSourceLocator(
```
theofidry: this is duplicated in several places, so maybe this should be refactored accordingly

src/Reflector.php
```diff
@@ -39,63 +39,344 @@
  */
 final class Reflector
 {
+    /**
+     * TODO: Migrate to Roave/BetterReflection once it supports constants
+     */
     private const KNOWN_INTERNAL_CONSTANTS = [
```
theofidry: @sebastianbergmann is there any constants from extensions PHPUnit relies on which are not in this list?

sebastianbergmann: Could be, don't know. Don't have a tool handy that lists all constants used.

tests/ReflectorTest.php
kukulich: You can simplify it with `BetterReflection::sourceStubber()`

theofidry: The configuration is a bit different though

theofidry:
@kukulich I can't investigate at the moment, but the tests are an order of magnitude slower. It takes about `2.82 seconds` to do `make tu` (run the unit tests) before and now about 5min

src/Reflector.php
theofidry: I'll try to have a look before merging that one. As long as the constants are not supported by BetterReflection it's a very brittle system so it would be nice it works fine at least for PHPUnit

kukulich:
@theofidry Yes, because it searches throw a lot of php files (stubs) now. I tried to optimize it as much as possible.

theofidry:
wasn't it doing so before as well?

kukulich:
@theofidry

Before: I can search files only in one folder because I know the extension (from internal reflection)
Now: I have to search throw all folders in https://github.com/JetBrains/phpstorm-stubs to find the function. 

So eg. searching `unknown` function is much slower now.

theofidry:
We are talking of a 100x slower right now... So maybe generating a lookup map (to be done later obviously) would be worthwhile? For some context, generating a scoped PHAR for [Box](https://github.com/humbug/box) _already_ takes about 80s (and it's using parallelisation).

kukulich:
Could you send me how to generate scoped PHAR for Box? I will try to find more optimizations.

theofidry:
- clone the repo
- install deps
- `bin/box compile`, for no parallelisation you can use the `--no-parallel` option

For the complete version it's `make compile` in which Box is being scoped 3 times in total (1 time to scope it, a second time to scope the Box PHAR with the scoped version and a third time to ensure the 2nd one works fine). But the command above should be enough for what you want to see

kukulich:
@theofidry Thanks.

Could you please try it now with last pushed version? I've just found bug in one optimization.

theofidry:
I think it actually made it worse :/

theofidry:
We'll see with the new build, but so far on Travis we are [17.2 minutes](https://travis-ci.org/humbug/php-scoper/jobs/530641524#L593) with this PR against [2.82 seconds](https://travis-ci.org/humbug/php-scoper/jobs/528493681#L579) on master

kukulich:
@theofidry Thanks. I will look at it.

kukulich:
@theofidry This should improve tests performence https://github.com/kukulich/php-scoper/commit/f49c9bfc0443dcf1da3a98b21d614baaf20f14d5 because the cache in `PhpStormStubsSourceStubber` should work now.

theofidry:
For the record, after your first optimization, it went down to [15.31 minutes](https://travis-ci.org/humbug/php-scoper/jobs/530695286#L593). Let's see where the other one brings us.

I'm however still concerned at the lookup speed. Your patch may be effective here since it enables caching for the whole test suite, but in practice like in Box, it's parallelised in workers so this caching strategy won't be nearly as effective. But maybe I'm just not optimistic enough. I'll try it later and see what's the result

theofidry:
Down to [22.79 seconds](https://travis-ci.org/humbug/php-scoper/jobs/530721318#L593) for the tests at least :) Still a x7 performance slowness, but already much better. Let's hope an actual scoping of a project is not too impacted

theofidry:
I think an easier project to work with to check this is Box requirement checker:

```
# master with PHP-Scoper 0.11.4
$ bin/box compile --no-parallel --working-dir requirement-checker
 // Memory usage: 13.55MB (peak: 20.77MB), time: <1sec

# With this branch
$ bin/box compile --no-parallel --working-dir requirement-checker
 // Memory usage: 16.89MB (peak: 29.03MB), time: 16secs
```

So it's at least 16 times slower so far. Some profiling with Blackfire would be needed for further inspections though

kukulich:
I've forcepushed new version. I hope it's better.

```
 # master with PHP-Scoper 0.11.4
 // Memory usage: 13.32MB (peak: 13.87MB), time: 1sec

 # With this branch
 // Memory usage: 16.23MB (peak: 17.52MB), time: 1sec
```

Please test.

theofidry:
>I've forcepushed new version.

I'm not sure to see it: did you force push on your repo or this one?

kukulich:
I've forcepushed https://github.com/Roave/BetterReflection/pull/470

theofidry:
Looks cool, it doesn't even show up in the diff: https://blackfire.io/profiles/compare/b69cc025-90ca-4f7c-b655-f0627bd5dfdf/graph

There is however an anonymous class serialization error that I need to figure out still

src/Reflector.php
theofidry: Outdated thanks to @kukulich we can leverage BetterReflector for which constants support has been added

theofidry:
Big thanks @kukulich :)

kukulich:
@theofidry You're welcome :)

PR #333, opened 2019-06-08
Update Reflector to account for missing symbols

theofidry:
Since it is unlikely that JetBrains' response time on its stubs library is satisfying in the short-term at least, I think it makes sense to take advantage of the current `Reflector` design to patch the stubs here in PHP-Scoper.

This PR introduce back a way to enrich the symbols taken from the stub map and add the following:

- symbols from the [`uv` extension](https://pecl.php.net/package/uv)
- symbols from the [`krakjoe/parallel` extension](https://github.com/krakjoe/parallel)
- symbols from the [`krakjoe/pcov` extension](https://github.com/krakjoe/pcov)
- symbols from the [`crypto` extension](https://pecl.php.net/package/crypto)
- the `sapi_windows_vt100_support()` function

/cc @sebastianbergmann this should make things easier for PHPUnit :)


sebastianbergmann:
Thanks. The PHAR is missing for the 0.12.1 release, though.

theofidry:
Fixed with 0.12.2

PR #363, opened 2019-12-13
Fix string resolution with leading backslas

### LINKED ISSUE #356: `ReflectionClassConstant ` has been reported as non-internal
See https://github.com/bobthecow/psysh/issues/581#issuecomment-560137900

theofidry:
Closes #356

### DIFF
--- a/src/PhpParser/NodeVisitor/StringScalarPrefixer.php
+++ b/src/PhpParser/NodeVisitor/StringScalarPrefixer.php
@@ -14,8 +14,14 @@ declare(strict_types=1);
 
 namespace Humbug\PhpScoper\PhpParser\NodeVisitor;
 
+use function array_key_exists;
+use function array_shift;
+use function array_values;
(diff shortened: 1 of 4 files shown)


src/PhpParser/NodeVisitor/StringScalarPrefixer.php
```diff
@@ -107,6 +108,8 @@ private function prefixStringScalar(String_ $string): String_
             return $string;
         }
 
+        $normalizedValue = '\\' === $string->value[0] ? substr($string->value, 1) : $string->value;
```
BackEndTea: Wouldn't an `ltrim` be cleaner here?

BackEndTea: e.g. `$normalizedValue  = ltrim($string->value, '\\');`

theofidry: Indeed, much better

PR #387, opened 2020-04-28
Fix coding style via StyleCI

peter279k:
As title. Fix coding style via StyleCI service.

### DIFF
--- a/bin/root-version.php
+++ b/bin/root-version.php
@@ -21,7 +21,8 @@ function get_last_tag_name(): string
         : "-H \"Authorization: token $gitHubToken\""
     ;
 
-    $lastReleaseEndpointContents = shell_exec(<<<BASH
+    $lastReleaseEndpointContents = shell_exec(
+        <<<BASH
 curl -s $headerOption https://api.github.com/repos/humbug/php-scoper/releases/latest
 BASH
     );
--- a/src/PhpParser/NodeVisitor/StringScalarPrefixer.php
+++ b/src/PhpParser/NodeVisitor/StringScalarPrefixer.php
@@ -131,13 +131,12 @@ final class StringScalarPrefixer extends NodeVisitorAbstract
         }
 
         if (false === (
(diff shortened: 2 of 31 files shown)


bin/root-version.php
```diff
@@ -34,23 +35,11 @@ function get_last_tag_name(): string
 
     if (JSON_ERROR_NONE !== json_last_error()) {
         // TODO: switch to safe json parsing in the future
-        throw new RuntimeException(
```
theofidry, line 37: This is one is not desired IMO. I think it's inherited from the Symfony style, but IMO it only hurts readability here

src/PhpParser/NodeVisitor/StringScalarPrefixer.php
```diff
@@ -131,12 +131,12 @@ private function prefixStringScalar(String_ $string): String_
...
         if (false === (
-                $parentNode instanceof Assign
+            $parentNode instanceof Assign
                 || $parentNode instanceof Param
                 || $parentNode instanceof Const_
                 || $parentNode instanceof PropertyProperty
                 || $parentNode instanceof Return_
-            )
+        )
         ) {
```
theofidry, line 140:
```suggestion
        )) {
```

peter279k: Fixed.

bin/root-version.php
peter279k: @theofidry, thanks for your reply.

Actually, I also agree with you, do you have any suggestion to let Style CI analyze the coding style successfully if revering these style changes?

theofidry: Should be configurable with `.styleci.yml`, as to what parameter exactly I do not know

peter279k: @theofidry, thanks for your reply.

I've disabled the `single_line_throw` setting on `.styleci.yml` setting to ignore this.

theofidry:
Thanks!

PR #390, opened 2020-05-02
Imrove PHPUnit fixtures

peter279k:
# Changed log
- According to the [PHPUnit fixtures reference](https://phpunit.readthedocs.io/en/8.5/fixtures.html?highlight=fixtures), it should be `protected setUp` method.


theofidry:
Thanks!

PR #398, opened 2020-07-12
Add compatibility with Composer 2

villfa:
*php-scoper* cannot be installed with Composer 2 (see below). [More details here](https://github.com/composer/package-versions-deprecated/blob/master/README.md).
```
$ git clone https://github.com/humbug/php-scoper && cd php-scoper
[...]
$ composer --version
Composer version 2.0.0-alpha2 2020-06-24 21:36:18
$ COMPOSER_ROOT_VERSION='0.13.99' composer install
Installing dependencies from lock file (including require-dev)
Verifying lock file contents can be installed on current platform.
Your lock file does not contain a compatible set of packages. Please run composer update.

  Problem 1
    - ocramius/package-versions is locked to version 1.4.2 and an update of this package was not requested.
    - ocramius/package-versions 1.4.2 requires composer-plugin-api ^1.0.0 -> found composer-plugin-api[2.0.0] but it does not match the constraint.
  Problem 2
    - ocramius/package-versions 1.4.2 requires composer-plugin-api ^1.0.0 -> found composer-plugin-api[2.0.0] but it does not match the constraint.
    - humbug/box 3.8.4 requires ocramius/package-versions ^1.4 -> satisfiable by ocramius/package-versions[1.4.2].
    - humbug/box is locked to version 3.8.4 and an update of this package was not requested.

Use the option --with-all-dependencies to allow upgrades, downgrades and removals for packages currently locked to specific versions.

ocramius/package-versions only provides support for Composer 2 in 1.8+, which requires PHP 7.4.
If you can not upgrade PHP you can require composer/package-versions-deprecated to resolve this with PHP 7.0+.

You are using a snapshot build of Composer 2, which some of your plugins seem to be incompatible with. Make sure you update your plugins or report an issue to them to ask them to support Composer 2. To work around this you can run Composer with --ignore-platform-req=composer-plugin-api, but this may result in broken plugins and bigger problems down the line.
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,8 @@ cache:
 
 env:
   global:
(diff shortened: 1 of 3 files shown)


villfa:
The build fails because of *humbug/box*. I have open another PR to fix it: https://github.com/humbug/box/pull/467

theofidry:
@villfa this PR can now be updated

villfa:
I plan to modify the Travis configuration to add a build with Composer 2. I don't know when it will be ready though.

villfa:
@theofidry the branch is up to date

theofidry:
Thanks @villfa!

Issue #399, opened 2020-07-23
[Scoper] Migrate to PHP-Parser name resolver


theofidry:
See https://github.com/sebastianbergmann/phpunit/issues/4382#issuecomment-662879951

sebastianbergmann:
@theofidry Sorry for bothering you, but do you know when you will have time to fix this? Thanks!

theofidry:
@sebastianbergmann checking it :) I think I found a quick solution for now.

@nikic I've been investigating to replace the PHP-Scoper's name resolver by the PHP-Parser one which I would love to do since it would remove a non-negligible amount of code from my side. I'm however blocked having some trouble getting around its API.

Given the following code:

```php
<?php

namespace Acme;

use Foo\Bar as X;

function ctr(X $x) {}
```

| description | PHP-Scoper NameResolver | PHP-Parser NameResolver |
|--|--|--|
| For the `Identifier` `ctr` | `Acme\ctr` | `ctr` |
| For the `Name` `X` (for the parameter) | `Foo\Bar` | `X` |

I'm have trouble to understand the first one and for the second one I am not sure to understand either: shouldn't the resolved name resolve the alias as well?

nikic:
There are two ways to use the name resolution functionality in PHP-Parser:

 1. Recommended: Register the NameResolver as a visitor (typically the first one). If you don't want to resolve names in place, you can set the `'replaceNodes' => false` option, which will only add a `resolvedName` attribute.
 2. Using the NameContext. This implements the actual name resolution logic, but requires you to feed it all the relevant information. In particular you need to call startNamespace() and addAlias() as necessary.

Or 3. which is to do 1. and then query the NameContext in the NameResolver if you additionally need to resolve names that are not part of the AST. The NameResolver visitor is going to take care of managing the context, so you only have to query it.

theofidry:
Thanks for the clarification! I somehow forgot to register it as a visitor...

I'll double check that again soonish, so I'm keeping this issue open. Meanwhile 0.13.3 should already contain a fix for you @sebastianbergmann

sebastianbergmann:
Thank you for working on this, on a Friday evening no less. Preliminary tests make me hopeful, I'll get back to you in case I run into problems.

theofidry:
👍 I'll try to do some PHP-Scoper/Box next week :)

theofidry:
Renamed the title since the bug itself has been fixed already

PR #400, opened 2020-08-07
WIP: Replace PHP-Scoper name resolver by the PHP-Parser one

theofidry:
Related to #399


theofidry:
Will be carried on in #506

PR #404, opened 2020-08-21
Fix InstalledPackagesScoper with Composer v2

villfa:
The format of `installed.json` has changed in Composer v2. See:

https://github.com/composer/composer/blob/56811b4c8fec9c2eda5bd47a2c079e121c668eb4/src/Composer/Repository/FilesystemRepository.php#L72-L75


theofidry:
thank you @villfa

PR #408, opened 2020-09-03
Add a "inverse-namespaces-whitelist" configuration option.

### LINKED ISSUE #406: [Scoper] Add support for inclusive namespace whitelisting
## Feature Request

Right now the `whitelist` section in the configuration file is _exclusive_: all the classes will be scoped except those whose namespace is in the whitelist.

It would be nice to have an option for the whitelist to be _inclusive_ instead, so that it works the reverse way: only the classes whose namespace is in the whitelist will be scoped, all others will be untouched.

I think the easier way to implement this would be to add a new boolean configuration option like `namespace_whitelist_is_inclusive`, defaulting to `false` for compatibility with the current behavior.

This is something I actually need for my job so I can implement it and submit a pull request. In fact I've already been experimenting with the code and seems definitely doable.

Konamiman:
Add a `inverse-namespaces-whitelist` option to the `scoper.inc.php`configuration file.

This is a boolean option that defaults to `false`. When set to `true`, the namespaces whitelist mechanism (`whitelist` configuration key) is reversed: only classes matching the namespaces listed will be prefixed, and all others will be left untouched.

This does not affect the whitelisting of global classes, constants or functions.

Closes #406.

### DIFF
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -46,6 +46,7 @@ final class Configuration
     private const WHITELIST_GLOBAL_CONSTANTS_KEYWORD = 'whitelist-global-constants';
     private const WHITELIST_GLOBAL_CLASSES_KEYWORD = 'whitelist-global-classes';
     private const WHITELIST_GLOBAL_FUNCTIONS_KEYWORD = 'whitelist-global-functions';
+    private const WHITELIST_IS_INVERTED_KEYWORD = 'inverse-namespaces-whitelist';
 
     private const KEYWORDS = [
         self::PREFIX_KEYWORD,
@@ -56,6 +57,7 @@ final class Configuration
         self::WHITELIST_GLOBAL_CONSTANTS_KEYWORD,
         self::WHITELIST_GLOBAL_CLASSES_KEYWORD,
         self::WHITELIST_GLOBAL_FUNCTIONS_KEYWORD,
+        self::WHITELIST_IS_INVERTED_KEYWORD,
     ];
 
     private $path;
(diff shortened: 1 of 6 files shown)


src/Configuration.php
```diff
@@ -384,7 +386,25 @@ private static function retrieveWhitelist(array $config): Whitelist
...
+                        self::WHITELIST_IS_INVERTED_KEYWORD,
+                        gettype($namespacesWhitelistIsInverted)
+                    )
+                );
+            }
+        }
+
+        $whitelist_instance = Whitelist::create($whitelistGlobalConstants, $whitelistGlobalClasses, $whitelistGlobalFunctions, ...$whitelist);
+        $whitelist_instance->setNamespacesWhitelistIsInverted($namespacesWhitelistIsInverted);
+        return $whitelist_instance;
```
theofidry, line 407: I would add a blank line between L406 & L407

src/Configuration.php
```diff
@@ -384,7 +386,25 @@ private static function retrieveWhitelist(array $config): Whitelist
...
+                    sprintf(
+                        'Expected %s to be a boolean, found "%s" instead.',
+                        self::WHITELIST_IS_INVERTED_KEYWORD,
+                        gettype($namespacesWhitelistIsInverted)
+                    )
+                );
+            }
+        }
+
+        $whitelist_instance = Whitelist::create($whitelistGlobalConstants, $whitelistGlobalClasses, $whitelistGlobalFunctions, ...$whitelist);
```
theofidry, line 405:
```suggestion
        $whitelistInstance = Whitelist::create($whitelistGlobalConstants, $whitelistGlobalClasses, $whitelistGlobalFunctions, ...$whitelist);
```

src/Whitelist.php
```diff
@@ -139,9 +140,17 @@ private function __construct(
         $this->constants = $constants;
         $this->namespaces = $namespaces;
         $this->patterns = $patterns;
+        $this->namespacesWhitelistIsInverted = false;
```
theofidry, line 143:
```suggestion
        $this->namespaceWhitelistIsInverted = false;
```

src/Whitelist.php
```diff
@@ -244,6 +260,16 @@ public function getRecordedWhitelistedClasses(): array
         return array_values($this->whitelistedClasses);
     }
 
+    public function getNamespacesWhitelistIsInverted(): bool
```
theofidry, line 263:
```suggestion
    public function isNamespaceWhitelistInverted(): bool
```

src/Whitelist.php
```diff
@@ -244,6 +260,16 @@ public function getRecordedWhitelistedClasses(): array
...
+    {
+        return $this->namespacesWhitelistIsInverted;
+    }
+
+    public function setNamespacesWhitelistIsInverted(bool $value): void
```
theofidry, line 268: is the setter necessary?

theofidry (approved):
Left a few nitpicks, but looks solid to me 👍

src/Whitelist.php
Konamiman: It is, since I didn't include this in the class constructor (the existing 8 arguments seemed too much already to add yet another one, and modifying unit tests would be a pain).

Konamiman:

theofidry:
This should now no longer be necessary via the expose/exclude config entries (part of 0.16)

PR #409, opened 2020-09-03
Add an "output-dir" configuration option

Konamiman:
Add an `output-dir` option to the configuration file.

This allows to specify the base output directory for the generated prefixed files in the configuration file, the same way as the
command line option of the same name already does. If both are present, the command line option takes precedence.

### DIFF
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -139,8 +139,7 @@ final class Configuration
 
         if (false === array_key_exists(self::OUTPUT_DIR_KEYWORD, $config)) {
             $outputDir = null;
-        }
-        else {
+        } else {
             $outputDir = $config[ self::OUTPUT_DIR_KEYWORD ];
         }
 
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -46,6 +46,7 @@ final class Configuration
     private const WHITELIST_GLOBAL_CONSTANTS_KEYWORD = 'whitelist-global-constants';
     private const WHITELIST_GLOBAL_CLASSES_KEYWORD = 'whitelist-global-classes';
     private const WHITELIST_GLOBAL_FUNCTIONS_KEYWORD = 'whitelist-global-functions';
+    private const OUTPUT_DIR_KEYWORD = 'output-dir';
 
     private const KEYWORDS = [
         self::PREFIX_KEYWORD,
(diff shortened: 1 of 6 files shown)


src/Configuration.php
```diff
@@ -134,7 +137,14 @@ public static function load(string $path = null, array $paths = []): self
         $filesFromPaths = self::retrieveFilesFromPaths($paths);
         $filesWithContents = self::retrieveFilesWithContents(chain($filesFromPaths, ...$finders));
 
-        return new self($path, $prefix, $filesWithContents, $patchers, $whitelist, $whitelistedFiles);
+        if (false === array_key_exists(self::OUTPUT_DIR_KEYWORD, $config)) {
+            $outputDir = null;
+        }
+        else {
```
theofidry, line 143: I would keep it on the same line:

```suggestion
       } else {
```

Same for the other occurrence

tests/Console/Command/AddPrefixCommandTest.php
```diff
@@ -138,23 +139,50 @@ public function test_get_version_menu(): void
         $this->fileSystemProphecy->isAbsolutePath(Argument::cetera())->shouldNotHaveBeenCalled();
     }
 
-    public function test_scope_the_given_paths(): void
+    /**
+     * @testWith [true, false, "output_dir_from_config"]
```
theofidry, line 143: Interesting, I was not aware of the `@testWith`. I would however prefer to move it in a data provider where it has the merit of being clear PHP code instead of comments

tests/Console/Command/AddPrefixCommandTest.php
```diff
@@ -138,23 +139,50 @@ public function test_get_version_menu(): void
...
+     *           [false, true, "output_dir_from_command_line"]
+     *           [true, true, "output_dir_from_command_line"]
+     *
+     * @param bool $use_output_dir_in_config_file True to use a configuration file with a 'output-dir' key
```
theofidry, line 147: the `@param`s can be skipped here, I think your variables are clear enough

tests/Console/Command/AddPrefixCommandTest.php
```diff
@@ -138,23 +139,50 @@ public function test_get_version_menu(): void
...
+     * @param bool $use_output_dir_in_command_line True to specify an output directory in command line
+     * @param string $expected_output_dir The actual output directory where the files will have been generated
+     */
+    public function test_scope_the_given_paths($use_output_dir_in_config_file, $use_output_dir_in_command_line, $expected_output_dir): void
```
theofidry, line 151: I would prefer camel case for the variables rather than snake case

theofidry (approved):
There is a few nitpicks, but LGTM otherwise 👌

Konamiman:

PR #411, opened 2020-09-10
Add an "on-existing-output-dir" configuration option

Konamiman:
This pull request fixes a bug and adds a new feature. The former is a prerequisite for the later.

#### Fixed bug

`validateOutputDir` in `AddPrefixCommand` asks the user for confirmation if the output directory already exists, however it ignores user input and proceeds with the scoping process regardless of what the user responds.

This has been fixed by having `validateOutputDir` return a boolean (which will be `false` if the user responds "no"); the `execute` function, in turn, will return immediately with an error code of 1 if `validateOutputDir` returns `false`.

#### New feature

Added a new `on-existing-output-dir` configuration option to decide what to do if the output directory already exists. Allowed values are 'ask' (default), 'overwrite' and 'abort'. The later is equivalent to the user responding "no" (so `execute` will return immediately with an error code of 1).

The `--force` command line option, if present, overrides the configuration option and works as 'overwrite' for compatibility.

### DIFF
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -46,6 +46,7 @@ final class Configuration
     private const WHITELIST_GLOBAL_CONSTANTS_KEYWORD = 'whitelist-global-constants';
     private const WHITELIST_GLOBAL_CLASSES_KEYWORD = 'whitelist-global-classes';
     private const WHITELIST_GLOBAL_FUNCTIONS_KEYWORD = 'whitelist-global-functions';
+    private const ON_EXISTING_OUTPUT_DIR_KEYWORD = 'on-existing-output-dir';
 
     private const KEYWORDS = [
         self::PREFIX_KEYWORD,
@@ -56,6 +57,7 @@ final class Configuration
         self::WHITELIST_GLOBAL_CONSTANTS_KEYWORD,
         self::WHITELIST_GLOBAL_CLASSES_KEYWORD,
         self::WHITELIST_GLOBAL_FUNCTIONS_KEYWORD,
+        self::ON_EXISTING_OUTPUT_DIR_KEYWORD
     ];
 
     private $path;
@@ -64,6 +66,7 @@ final class Configuration
     private $patchers;
     private $whitelist;
     private $whitelistedFiles;
+    private $onExistingOutputDir;
 
     /**
      * @param string|null $path  Absolute path to the configuration file.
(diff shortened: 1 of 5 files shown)


src/Configuration.php
```diff
@@ -548,4 +561,42 @@ static function (array $files, SplFileInfo $fileInfo): array {
             []
         );
     }
+
+    /**
+     * @param array $config
```
theofidry, line 566: the phpdoc can be omitted here since redundant with the signature

src/Configuration.php
```diff
@@ -548,4 +561,42 @@ static function (array $files, SplFileInfo $fileInfo): array {
...
+                sprintf(
+                    'Expected %s to be a string, found "%s" instead.',
+                    self::ON_EXISTING_OUTPUT_DIR_KEYWORD,
+                    gettype( $onExistingOutputDir )
+                )
+            );
+        }
+
+        $allowedOnExistingOutputDirValues = ['ask', 'overwrite', 'abort'];
+        if(!in_array($onExistingOutputDir, $allowedOnExistingOutputDirValues)) {
```
theofidry, line 589:
```suggestion
        if(!in_array($onExistingOutputDir, $allowedOnExistingOutputDirValues, true)) {
```

src/Configuration.php
```diff
@@ -548,4 +561,42 @@ static function (array $files, SplFileInfo $fileInfo): array {
...
+            throw new InvalidArgumentException(
+                sprintf(
+                    'Found value of "%s" for %s; allowed values are %s.',
```
theofidry, line 592:
```suggestion
                    'Expected one of "%s" for "%s". Got "%s"',
```

Careful you need to rearrange the order the parameters with this suggestion

src/Configuration.php
```diff
@@ -548,4 +561,42 @@ static function (array $files, SplFileInfo $fileInfo): array {
...
+                    $onExistingOutputDir,
+                    self::ON_EXISTING_OUTPUT_DIR_KEYWORD,
+                    implode(', ', $allowedOnExistingOutputDirValues)
```
theofidry, line 595:
```suggestion
                    implode('", "', $allowedOnExistingOutputDirValues)
```

src/Console/Command/AddPrefixCommand.php
```diff
@@ -301,7 +304,13 @@ static function (string $path) use ($cwd, $fileSystem) {
         $input->setArgument(self::PATH_ARG, $paths);
     }
 
-    private function validateOutputDir(InputInterface $input, OutputStyle $io): void
+    /**
+     * @param InputInterface $input
```
theofidry, line 308: the `@param` can be omitted here since redundant with the signature

src/Configuration.php
theofidry, line 589: if `in_array()` is not included in the list of imports I would add it (`use function in_array;`) . Likewise for the others

theofidry (approved):
LGTM. There is a few nitpicks left as well as the CI that is still unhappy (don't forget to merge the latests changes it might help) but looks good otherwise

Konamiman:

theofidry:
I unfortunately was too slow to review and merge this. I cannot rebase it now and it would be quite a bit of work. I however could at least fix the bug reported here. Since the PR has been stale for 2 years I'll be closing it (it is entirely my fault though)

PR #417, opened 2020-10-09
Also check for interfaces and traits in existence checks

### LINKED ISSUE #403: scoper-autoload.php using class_exists for scoped interfaces
## Bug report

| Question    | Answer
| Box version | 0.13.3
| PHP version | 7.2.29

scoped output of `vendor/symfony/polyfill-php70/Resources/stubs/SessionUpdateTimestampHandlerInterface.php`

```php
<?php

namespace MyNamespace;

interface SessionUpdateTimestampHandlerInterface
{
    /**
     * Checks if a session identifier already exists or not.
     *
     * @param string $key
     *
     * @return bool
     */
    public function validateId($key);
    /**
     * Updates the timestamp of a session when its data didn't change.
     *
     * @param string $key
     * @param string $val
     *
     * @return bool
     */
    public function updateTimestamp($key, $val);
}
\class_alias('MyNamespace\\SessionUpdateTimestampHandlerInterface', 'SessionUpdateTimestampHandlerInterface', \false);
```

`scoper-autoload.php` contains the following snippet:
```php
if (!class_exists('SessionUpdateTimestampHandlerInterface', false)) {
    class_exists('MyNamespace\SessionUpdateTimestampHandlerInterface');
}
```
The above results in an error:
```
Cannot declare interface SessionUpdateTimestampHandlerInterface, because the name is already in use

vendor/symfony/polyfill-php70/Resources/stubs/SessionUpdateTimestampHandlerInterface.php:25
class_alias()
vendor/symfony/polyfill-php70/Resources/stubs/SessionUpdateTimestampHandlerInterface.php:25
Composer\A\includeFile()
vendor/composer/ClassLoader.php:322
Composer\A\ClassLoader->loadClass()
Unknown location
spl_autoload_call()
Unknown location
class_exists('MyNamespace\SessionUpdateTimestampHandlerInterface')
vendor/scoper-autoload.php:25
```

If I change the `scoper-autoload.php` snippet above to the following instead, the error is gone:
```php
if (!interface_exists('SessionUpdateTimestampHandlerInterface', false)) {
    interface_exists('MyNamespace\SessionUpdateTimestampHandlerInterface');
}
```

Is there something I am missing to make sure `scoper-autoload.php` correctly uses `interface_exists` for Interfaces?

discordier:
Only performing `class_exists()` is not sufficient as interfaces and
traits are different kind of autoloadable things.
Therefore we have to test that none of these exist prior trying to
autoload an alias them.

Fixes #403

### DIFF
--- a/fixtures/set011/composer.json
+++ b/fixtures/set011/composer.json
@@ -3,6 +3,7 @@
         "bin/greet.php"
     ],
     "autoload": {
+        "classmap": ["polyfill"],
         "psr-4": {
             "Set011\\": "src/"
         }
--- a/fixtures/set011/polyfill/Iterator.php
+++ b/fixtures/set011/polyfill/Iterator.php
@@ -1,0 +1,18 @@
+<?php
+
+/**
+ * Let's polyfill the iterator interface.
+ *
+ * We use it as example to alias an interface into the root namespace.
+ * It will never get loaded for real as it exists on all PHP versions.
+ *
+ * @link https://php.net/manual/en/class.iterator.php.
+ * @link https://github.com/humbug/php-scoper/issues/403
+ */
+interface Iterator extends Traversable {
+    public function current();
+    public function next(): void;
+    public function key();
+    public function valid(): bool;
+    public function rewind(): void;
+}
--- a/src/Autoload/ScoperAutoloadGenerator.php
+++ b/src/Autoload/ScoperAutoloadGenerator.php
@@ -113,8 +113,8 @@ PHP;
 
                 return sprintf(
                     <<<'PHP'
-if (!class_exists('%s', false)) {
-    class_exists('%s');
+if (!class_exists('%1$s', false) && !interface_exists('%1$s', false) && !trait_exists('%1$s', false)) {
+    spl_autoload_call('%2$s');
 }
 PHP
                     ,
--- a/tests/Autoload/ScoperAutoloadGeneratorTest.php
+++ b/tests/Autoload/ScoperAutoloadGeneratorTest.php
@@ -187,8 +187,8 @@ $loader = require_once __DIR__.'/autoload.php';
 
 // Aliases for the whitelisted classes. For more information see:
 // https://github.com/humbug/php-scoper/blob/master/README.md#class-whitelisting
-if (!class_exists('A\Foo', false)) {
-    class_exists('Humbug\A\Foo');
+if (!class_exists('A\Foo', false) && !interface_exists('A\Foo', false) && !trait_exists('A\Foo', false)) {
+    spl_autoload_call('Humbug\A\Foo');
 }
 
 return $loader;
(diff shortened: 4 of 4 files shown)


theofidry (commented):
Thanks a lot for the PR! Could you try to expand the `fixtures/set011` (which is an e2e test for thee whitelist feature)? Alternatively another set could be added

discordier:
I wonder how to integrate it.

I think I'll try to "polyfill" the interface `Iterator` that exists in PHP.
This should never be tried to be loaded but added to the whitelist.

What do you think?

However, I won't have time until monday to work on it.

discordier:
Test added.

However, we might have a race condition in our entire class aliasing when any whitelisted class depends on another white listed class, as the list of autoload calls is not ordered by dependency.
Solving this is not scope of this PR but I wanted to mention it anyway.

Imagine the following:
1. We polyfill a global interface `Iterator` and add it to the `classmap` in `composer.json`
2. We use this polyfill in a white listed class implementation `DictionaryIterator`:
```php
class DictionaryIterator implements \Iterator {
/* ... */
}
```

We will get the following in `autoload.php`:
```php
// Aliases for the whitelisted classes. For more information see:
// https://github.com/humbug/php-scoper/blob/master/README.md#class-whitelisting
if (!class_exists('Set011\DictionaryIterator', false) && !interface_exists('Set011\DictionaryIterator', false) && !trait_exists('Set011\DictionaryIterator', false)) {
    spl_autoload_call('_HumbugBoxbfd54d19c088\Set011\DictionaryIterator');
}
if (!class_exists('Set011\Dictionary', false) && !interface_exists('Set011\Dictionary', false) && !trait_exists('Set011\Dictionary', false)) {
    spl_autoload_call('_HumbugBoxbfd54d19c088\Set011\Dictionary');
}
if (!class_exists('Iterator', false) && !interface_exists('Iterator', false) && !trait_exists('Iterator', false)) {
    spl_autoload_call('_HumbugBoxbfd54d19c088\Iterator');
}
```

We are now loading the depending class `Set011\DictionaryIterator` before(!) the interface `Iterator` and hence make it non loadable.

To solve this, we would need to move the class aliasing out of the class declarations into the autoloader chain (via custom autoload function would be the best I think).

I'll try to give it a go over the next days but wanted to solve this immediate problem first (because I need it in a project of mine).

So this is good to merge from my side.

theofidry:
Thank you @discordier 

It's indeed a tricky problem 🤔

discordier:
After checking the output more closely for above mentioned case, we should be save for white listed classes, as they should only extend the scoped class alias. Sorry for the noise on this one. I still think we might want to give the custom autoload function a shot, if not only for not loading all the whitelisted elements upon start up but only when requested.

As said before, I'll try to twiddle with it over the next days.

Thanks for merging.

theofidry:
>I still think we might want to give the custom autoload function a shot, if not only for not loading all the whitelisted elements upon start up but only when requested.

It's a reasonable way IMO, although it will only apply to classes since there is no autoloading for functions/constants

PR #418, opened 2020-10-21
Update Reflector.php with constants from PHP 8.0 (and 7.4)

sanmai:
As per https://github.com/infection/infection/pull/1354#discussion_r509246724


theofidry:
Thank you @sanmai

Issue #419, opened 2020-10-24
No PHARs for latest releases


sebastianbergmann:
0.13.6 and 0.13.7 have no PHAR.

smoench:
It is fixed with #422

PR #421, opened 2020-10-28
Enabling GitHub Actions as CI + phar uploading on releases

smoench:
I read about the issue #419 that phar files were not uploaded for the latest releases. This might be an issue with travis so I was wondering if GitHub Actions could be a solution.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,88 @@
+name: Build
+
+on:
+    push:
+        branches: [master]
+    pull_request:
+    release:
+        types: [created]
+
+jobs:
+    tests:
+        runs-on: ubuntu-latest
+        name: Build and test
+        strategy:
+            matrix:
+                php: ["7.2", "7.4"]
+                composer-version: ["1"]
+                coverage: ["none"]
+                publish-phar: [false]
+                include:
+                    -   php: "7.3"
+                        coverage: "pcov"
+                        publish-phar: true
+                    -   php: "7.4"
+                        composer-version: "2"
+
+        steps:
+            -   uses: actions/checkout@v2
+
+            -   name: Setup PHP
+                uses: shivammathur/setup-php@v2
+                with:
+                    php-version: "${{ matrix.php }}"
+                    ini-values: "phar.readonly=0"
+                    tools: "composer:v${{ matrix.composer-version }}"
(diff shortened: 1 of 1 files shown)


theofidry:
I'd be happy to switch to GitHub Actions; I just did not have the energy to try it. I'll check this out ASAP thanks!

smoench:
I also can contribute the switch to GitHub Action for box or if you want to try it I can support you :-)

.github/workflows/build.yml
```diff
@@ -0,0 +1,88 @@
...
+
+    publish-phar:
+        runs-on: ubuntu-latest
+        name: Publish the PHAR
+        needs: tests
+        if: github.event_name == 'release'
+        steps:
+            -   uses: actions/download-artifact@v1
+                with:
+                    name: php-scoper.phar
```
theofidry, line 78: I think the path is `bin/php-scoper.phar`

.github/workflows/build.yml
```diff
@@ -0,0 +1,88 @@
...
+    release:
+        types: [created]
+
+jobs:
+    tests:
+        runs-on: ubuntu-latest
+        name: Build and test
+        strategy:
+            matrix:
+                php: ["7.2", "7.4"]
```
theofidry, line 16: Could you bump it to 7.3? I think it's time to drop 7.2; Maybe we can include 8.0 as well

theofidry (approved):
LGTM overall 👍 

How come it is not being displayed on the build, is there something I need to enable besides that?

theofidry:
>I also can contribute the switch to GitHub Action for box or if you want to try it I can support you :-)

That would be very welcomed!

smoench:
> LGTM overall 👍 
> 
> How come it is not being displayed on the build, is there something I need to enable?

It needs once in a origin branch to be detected. You can merge in the master or create an extra branch which I prefer. Then I can add PHP 8.0 support and possible errors.

.github/workflows/build.yml
smoench: No, it is the name of the artifact defined in line 67

smoench: No, it is the name of the artifact defined in line 67

theofidry:
I see, I'll merge it then and you can address the comments afterwards which would also give the benefit of seeing the whole pipeline!

PR #422, opened 2020-10-29
Fix GitHub Actions

smoench:


smoench:
@theofidry tests are working with github actions so far but `box compile` fails with:
```
 The tag or commit hash could not be retrieved from "/home/runner/work/php-s  
  coper/php-scoper": fatal: No names found, cannot describe anything.        
```

.github/workflows/build.yml
```diff
@@ -19,11 +19,15 @@ jobs:
...
                 include:
                     -   php: "7.3"
+                        composer-version: "1"
                         coverage: "pcov"
                         publish-phar: true
                     -   php: "7.4"
                         composer-version: "2"
 
+        env:
+            COMPOSER_ROOT_VERSION: '0.13.99'
```
theofidry, line 29: is there any way this is loaded from the `.composer-root-version` instead? This would make it slightly easier to maintain since only one place would need to be updated

smoench: The source command didn't work because this how an env var should be set https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable

theofidry: Maybe that's still possible then?

```
source .composer-root-version
echo "COMPOSER_ROOT_VERSION =${COMPOSER_ROOT_VERSION}" >> $GITHUB_ENV
```

.github/workflows/build.yml
```diff
@@ -19,13 +19,19 @@ jobs:
...
+
         steps:
             -   uses: actions/checkout@v2
+                with:
+                    fetch-depth: 0
```
smoench, line 34: With fetching all the history the `box compile` works

.github/workflows/build.yml
smoench: Yes, this works 🎉

.github/workflows/build.yml
theofidry: It can be because composer uses git for checking the installed versions or something which is used in Box

smoench: It is needed only for the phar build so it can be configurable via the matrix section

theofidry: 👍

smoench: Done

smoench:
Ready for merge. Only the job "Publish the PHAR" needs to be tested. Wanna try with a new release for php-scoper? 😬 
I think travis needs to be deactivated before.

theofidry:
Let's test the PHAR publishing then :)

theofidry:
Let's see how it goes: https://github.com/humbug/php-scoper/releases/tag/0.13.8

smoench:
It worked

theofidry:
Nice! Would it be possible to do the same for Box?

smoench:
Yes, sure. Can you copy paste this workflow file in an extra branch for box. So I can create a PR against it with a running CI?

theofidry:
Will do tonight

theofidry:
I pushed the `feature/ga` on Box

PR #423, opened 2020-10-30
Correct function in docs

XedinUnknown:
Documentation says that `class_exists()` is used to autoload the aliased class, but `class_alias()` is used in the example.


theofidry:
Thank you @XedinUnknown

PR #425, opened 2020-10-30
Add versions to hot-patched symbols

smoench:
fixes #420

PR #426, opened 2020-11-01
Bump dependencies

theofidry:


theofidry:
Hm apparently the new version of Infection requires PHP 7.4... WDYT of bumping PHP-Scoper as well to that version @villfa @smoench?

smoench:
Let us fix #425 first :-)

theofidry:
Actually the Symfony build is failing due to that so indeed it will wait for it :P

villfa:
> Hm apparently the new version of Infection requires PHP 7.4... WDYT of bumping PHP-Scoper as well to that version?

I don't have a strong opinion about that.
I'm just a bit worried about the inconvenience it may cause to the users of *php-scoper* and *box* (since *box* will have to follow the move), especially there is no docker image provided.

smoench:
Sorry, @theofidry you are doing here a bit too much so I created #428 and #429 😬

PR #428, opened 2020-11-04
update composer/package-versions-deprecated

smoench:
This PR updates `composer/package-versions-deprecated` to the latest version `1.11.99`.


theofidry:
Thank you @smoench

PR #432, opened 2020-11-05
force use statements for global function, constants and classes if they are not colliding

smoench:
Alphabetically sorted uses are grouped in that order `classes`, `functions` and `constants` https://github.com/slevomat/coding-standard#slevomatcodingstandardnamespacesalphabeticallysorteduses-.
This is the default behavior of PSR-12, so StyleCI is wrong ;-) Probably there is a config option.


theofidry:
👍 I disabled StyleCI

PR #437, opened 2020-11-13
Revert "Add versions to hot-patched symbols" but keep fixed Lexer

smoench:
I was wrong with the Reflector approach. It will prevent us scoping the php-parser.

It is related to https://github.com/box-project/box/pull/499

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,7 +44,7 @@ cs-check: $(CODE_SNIFFER)
 	$(PHPNOGC) $(CODE_SNIFFER)
 
 .PHONY: phpstan
-PHPSTAN=bin/phpstan
+PHPSTAN=vendor-bin/phpstan/vendor/bin/phpstan
 phpstan: ## Runs PHPStan
 phpstan: $(PHPSTAN)
 	$(PHPNOGC) $(PHPSTAN) analyze src --level max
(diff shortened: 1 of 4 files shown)


smoench:
Can someone have a look why phpstan fails? Something happened with the Makefile? It worked when I had contributed it.

theofidry:
Could you elaborate a bit on what's going on? I'm not sure to understand the back & forth here...

smoench:
php-parser uses and defines internally T_FN, T_MATCH and so forth constants. In case your scoping php-parser those constants (e.g. `Humbug123\T_FN`) would be scoped but php-parser still defines them on runtime without the scoped namespace. Then box will fail as `Humbug123\T_FN` constant is not defined.

theofidry:
>Then box will fail as Humbug123\T_FN constant is not defined.

Why is that? If PHP-Parser still relies on defining them at runtime won't both (scoped & non-scoped symbols) be found?

smoench:
They are defined on runtime when they are not presented so they are stored as strings -> https://github.com/nikic/PHP-Parser/blob/master/lib/PhpParser/Lexer.php#L413

theofidry:
Sorry for the delay @smoench, thanks for the PR

PR #438, opened 2020-11-16
Move jetbrains/phpstorm-stubs to a dev-dependency.

herregroen:
I'm not sure why jetbrains/phpstorm-stubs was a dependency in the first place but it's currently causing us issues as `dev-master` no longer seems to exist on composer.org.

Given that these are stubs for use in a specific IDE I believe these should be a dev dependency and not a normal dependency.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,6 @@
     "require": {
         "php": "^7.3",
         "composer/package-versions-deprecated": "^1.8",
-        "jetbrains/phpstorm-stubs": "dev-master",
         "nikic/php-parser": "^4.0",
         "symfony/console": "^3.2 || ^4.0",
         "symfony/filesystem": "^3.2 || ^4.0",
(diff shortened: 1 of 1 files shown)


smoench:
This might be issue with jetbrains/phpstorm-stubs itself https://github.com/JetBrains/phpstorm-stubs/pull/955

jetbrains/phpstorm-stubs is being used for detecting internal classes, constants and functions so they don't get prefixed.

theofidry:
>Given that these are stubs for use in a specific IDE I believe these should be a dev dependency and not a normal dependency.

They actually are a non-dev dependency since PHP-Scoper relies on those stubs to determine if a given symbol is internal or not

smoench:
Can be closed as https://github.com/JetBrains/phpstorm-stubs/pull/955 has been fixed

PR #445, opened 2020-11-29
Improve file whitelist performances

theofidry:
Extracted from #410


theofidry:
/cc @Konamiman

I've been tinkering with a few things in PHP-Scoper & Box. Although the path to get there was very different, in the end I think this is pretty much what you have in #410. A few differences

- This is strictly about the file whitelisting performance (in your PR there is a few other changes which would be nice to have as well but in a dedicated PR)
- It does not remove the `FileWhitelistScoper`: it is used in Box

PR #447, opened 2020-12-03
Parentheses gets removed

smoench:
see https://github.com/smoench/php-scoper/pull/1#issuecomment-737534466


smoench:
@villfa It is an issue with nikic/php-parser but it is already fixed on master. We need to wait for the next release I guess.

smoench:
A new release of php-parser has been released fixing this issue. Should we keep the test case?

villfa:
> Should we keep the test case?

Probably not since it basically tests the behaviour of PHP-Parser, and PHP-Parser has already its own test for that case now.

PR #450, opened 2020-12-23
update box to version 3.11 and bump composer root version to 0.14.99

smoench:
waits for https://github.com/box-project/box/pull/517


theofidry:
Danke!

Issue #440, opened 2020-11-25
[Scoper] Add polyfill support


hustlahusky:
## Bug report

| Question    | Answer
| ------------| ---------------
| Box version | 0.13.9
| PHP version | 7.3.22
| Platform with version | Alpine (docker)
| Github Repo | -

Function aliasing described in [docs](https://github.com/humbug/php-scoper#functions-whitelisting) seems not working with symfony polyfills functions. File scoped by namespace, but not `function_exist` check, also no aliases in `scoper-autoload.php`.

Can this caused because they all are oneliners? For example, `dd` function from `symfony/var-dumper` prefixed and aliased as expected.

<details>
 <summary>Original vendor/symfony/polyfill-php80/bootstrap.php</summary>
 
 ```php
<?php

/*
 * This file is part of the Symfony package.
 *
 * (c) Fabien Potencier <fabien@symfony.com>
 *
 * For the full copyright and license information, please view the LICENSE
 * file that was distributed with this source code.
 */

use Symfony\Polyfill\Php80 as p;

if (PHP_VERSION_ID >= 80000) {
    return;
}

if (!defined('FILTER_VALIDATE_BOOL') && defined('FILTER_VALIDATE_BOOLEAN')) {
    define('FILTER_VALIDATE_BOOL', FILTER_VALIDATE_BOOLEAN);
}

if (!function_exists('fdiv')) {
    function fdiv(float $num1, float $num2): float { return p\Php80::fdiv($num1, $num2); }
}
if (!function_exists('preg_last_error_msg')) {
    function preg_last_error_msg(): string { return p\Php80::preg_last_error_msg(); }
}
if (!function_exists('str_contains')) {
    function str_contains(string $haystack, string $needle): bool { return p\Php80::str_contains($haystack, $needle); }
}
if (!function_exists('str_starts_with')) {
    function str_starts_with(string $haystack, string $needle): bool { return p\Php80::str_starts_with($haystack, $needle); }
}
if (!function_exists('str_ends_with')) {
    function str_ends_with(string $haystack, string $needle): bool { return p\Php80::str_ends_with($haystack, $needle); }
}
if (!function_exists('get_debug_type')) {
    function get_debug_type($value): string { return p\Php80::get_debug_type($value); }
}
if (!function_exists('get_resource_id')) {
    function get_resource_id($res): int { return p\Php80::get_resource_id($res); }
}
 ```
</details>

<details>
 <summary>Prefixed one</summary>
 
 ```php
<?php

[...]

PR #457, opened 2021-02-25
Not adding prefixes to names after use statements

### LINKED ISSUE #456: Use directive ignored for prefixes
## Bug report

| Question    | Answer
| PHP-Scoper version | 0.14.0
| PHP version | 7.4.15
| Platform with version | Debian testing
| Github Repo | Not public

Hi,
I'm trying to use php-scoper on the Google SDK and I am getting prefixes for classes that have already been brought into scope with the `use` directive. I tried shortening it a bit, but here's my example:

Source file:
```php
<?php
namespace Google\AccessToken;

use Google\Auth\HttpHandler\HttpHandlerFactory;
use Google\Client;
use GuzzleHttp\ClientInterface;
use GuzzleHttp\Psr7;
use GuzzleHttp\Psr7\Request;

/**
 * Wrapper around Google Access Tokens which provides convenience functions
 *
 */
class Revoke
{
  /**
   * @var ClientInterface The http client
   */
  private $http;

  /**
   * Instantiates the class, but does not initiate the login flow, leaving it
   * to the discretion of the caller.
   */
  public function __construct(ClientInterface $http = null)
  {
    $this->http = $http;
  }
}
```

Output:
```php
<?php
namespace Sgdg\Vendor\Google\AccessToken;

use Sgdg\Vendor\Google\Auth\HttpHandler\HttpHandlerFactory;
use Sgdg\Vendor\Google\Client;
use Sgdg\Vendor\GuzzleHttp\ClientInterface;
use Sgdg\Vendor\GuzzleHttp\Psr7;
use Sgdg\Vendor\GuzzleHttp\Psr7\Request;
/**
 * Wrapper around Google Access Tokens which provides convenience functions
 *
 */
class Revoke
{
    /**
     * @var ClientInterface The http client
     */
    private $http;
    /**
     * Instantiates the class, but does not initiate the login flow, leaving it
     * to the discretion of the caller.
     */
    public function __construct(\Sgdg\Vendor\GuzzleHttp\ClientInterface $http = null)
    {
        $this->http = $http;
    }
}
```

As you can see, in the constructor, the `ClientInterface` class is prefixed, but AFAIK that is not necessary as it has already been brought into scope with the `use` directive.

marekdedic:
Closes #456 

I am new to the codebase and it isn't exactly simple, so please check if this breaks anything...


marekdedic:
Ok, so I'm going through the failing unit tests and fixing them, but there are a few cases that need comments:
- `e2e_026`: I think it was broken before, but the test didn't catch it - the `use const` statement wasn't prefixed (but should've been IMO)

marekdedic:
I think this is ready for review. Please review the changes. I'd focus on stuff around whitelisting as I am still a little fuzzy about the details of how it's supposed to work. And also e2e test 32 where I had to put in a special workaround - I'm not really sure whether there isn't done bigger underlying issue I'm missing completely or is it's just caused by using php-scoper on itself.


BTW all the tests really helped to get this into shape, thanks for being diligent about them!

theofidry:
Sorry for the tardiness @marekdedic, I'll try to check it out ASAP

marekdedic:
Thanks, don't worry, I know what it's like with volunteer-run projects :slightly_smiling_face:

src/PhpParser/NodeVisitor/NameStmtPrefixer.php
```diff
@@ -204,4 +226,14 @@ private function prefixName(Name $name): Node
             $resolvedName->getAttributes()
         );
     }
+
+    private static function array_starts_with($arr, $prefix): bool
+    {
+        for ($i = 0; $i < count($prefix); ++$i) {
```
theofidry, line 232: be aware that the count here is executed every single time which is not ideal

src/PhpParser/NodeVisitor/NameStmtPrefixer.php
```diff
@@ -146,6 +155,19 @@ private function prefixName(Name $name): Node
...
+        if (
+            $useStatement !== null and
+            self::array_starts_with($resolvedName->parts, $useStatement->parts) and
+            !($parentNode instanceof ConstFetch and ($this->whitelist->isGlobalWhitelistedConstant($resolvedName->toString()) or $this->whitelist->isSymbolWhitelisted($resolvedName->toString(), true))) and
+            !($useStatement->getAttribute('parent') and $useStatement->getAttribute('parent')->alias !== null and $this->whitelist->isSymbolWhitelisted($useStatement->toString())) and
+            !($name instanceof FullyQualified) and
+            $resolvedName->parts !== ['Isolated', 'Symfony', 'Component', 'Finder', 'Finder']
+        ) {
+            return $name;
+        }
```
theofidry, line 169: There is several things I'm not too fond of here:

- the `and` (see https://github.com/kalessil/phpinspectionsea/blob/master/docs/control-flow.md#non-optimal-if-conditions)
- the long condition

the later though is hardly specific from your addition I'm already unpleased with the existing code in that regard. I'd just change a bit the indentation to make it a bit easier to follow:

```suggestion

        if (
            $useStatement !== null
            && !($name instanceof FullyQualified)
            && self::array_starts_with($resolvedName->parts, $useStatement->parts)
            && !(
                $parentNode instanceof ConstFetch
                && (
                    $this->whitelist->isGlobalWhitelistedConstant($resolvedName->toString())
                    || $this->whitelist->isSymbolWhitelisted($resolvedName->toString(), true)
                )
            )
            && !(
                $useStatement->getAttribute('parent')
                && $useStatement->getAttribute('parent')->alias !== null
                && $this->whitelist->isSymbolWhitelisted($useStatement->toString())
            )
            && $resolvedName->parts !== ['Isolated', 'Symfony', 'Component', 'Finder', 'Finder']
        ) {
            return $name;
        }
```

theofidry (approved):
I did not bother with a Blackfire profile but I can't see any notable speed difference so LGTM, thanks a lot for the impressive work!

>I'd focus on stuff around whitelisting as I am still a little fuzzy about the details of how it's supposed to work.

Regarding the whitelisting it's something I would like to change as the terminology is both deceptive and inappropriate (without going too far into diversity stuff I am not too fond of the whitelist/blacklist terminology).

Ideally I would like to come with terms that can properly express that in one case the symbols are left unchanged (e.g. for "whitelist-files" the files are not changed at all) and in the other they are still scoped but an alias is added to allow using its non-scoped symbol (more details on this in https://github.com/humbug/php-scoper#implementation-insights)

>And also e2e test 32 where I had to put in a special workaround - I'm not really sure whether there isn't done bigger underlying issue I'm missing completely or is it's just caused by using php-scoper on itself.

I am a bit confused by that one as well. In case this is unclear:

- PHP-Scoper ships with a Finder which is scoped
- Since PHP-Scoper loads directly the PHP config file, although there is other ways, in order to leverage the Finder without requiring the user to autoload special code PHP-Scoper exposes its finder (see `bin/php-scoper`:)

```php
// Exposes the finder used by PHP-Scoper PHAR to allow its usage in the configuration file.
if (false === class_exists(IsolatedFinder::class)) {
    class_alias(Finder::class, IsolatedFinder::class);
}
```

However the above would still be scoped and to avoid that I am relying on a patcher in the config:

```
        static function (string $filePath, string $prefix, string $contents): string {
            if ('bin/php-scoper' !== $filePath) {
                return $contents;
            }

            return str_replace(
                '\\'.$prefix.'\Isolated\Symfony\Component\Finder\Finder::class',
                '\Isolated\Symfony\Component\Finder\Finder::class',
                $contents
            );
        },
```

The fixture 032 is to make sure this process works, i.e. that the isolated finder is exposed and usable.

>BTW all the tests really helped to get this into shape, thanks for being diligent about them!

Glad they helped out, they are annoying to write and a bit slow but hopefully they pay for themselves

src/PhpParser/NodeVisitor/NameStmtPrefixer.php
marekdedic: Ha, forgot I was in one of the languages that does this, thanks! :)

marekdedic:
I agree that the term "whitelisted" can be really confusing for this use. If I may, I'd suggest ignored/excluded for files that are completely untouched. For the other use, maybe something like "add-alias-to-original"?

I fixed both your suggestions, thank you for them. Is there anything more to do on my part? (e.g. regarding the Finder or whitelisting?)

theofidry:
>I'd suggest ignored/excluded for files that are completely untouched. For the other use, maybe something like "add-alias-to-original"?

I think those are good suggestions

>I fixed both your suggestions, thank you for them. Is there anything more to do on my part? (e.g. regarding the Finder or whitelisting?)

I think the Finder can wait; it's all good I just wanted to do a patch release before merging this :)

theofidry:
Thanks a lot @marekdedic

TomasVotruba:
Thank you @marekdedic :+1:

TomasVotruba:
@theofidry Hi Theo, this feature would be very helpful with solving https://github.com/humbug/php-scoper/issues/440#issuecomment-795160132

Too bad patch release was before this and not after :D
Any ETA on phar release?

TomasVotruba:
@theofidry Thank you for such a fast action :clap: 
I'm gonna try it on Rector right now

marekdedic:
Hi @TomasVotruba, I am a bit lost in your code and cannot reproduce the issue. Could you please narrow it down to some minimal reproducible example?

Thanks.

PR #461, opened 2021-03-07
Updated dependencies to allow for "minimum-stability: stable" installs

marekdedic:


theofidry:
is this really needed? The stubs are not really being tagged consistently with the changes so being on dev-master for it is preferable

marekdedic:
Well, it's a significant "barrier to entry" since most composer projects use `minimum-stability: stable` and composer refuses to let them install php-scoper then. I know there is a workaround with composer-bin-plugin, but then you have to manage that in the CI and all contributors need to do it as well and while all of this can be overcome, I think it would be much easier if you wouldn't have to...

The dependency in question is weird, that's true. But given that they release cca 3x a year, is that really an issue? I don't know if you use automated dependency updating, but with that, that would seem fine to me...

theofidry:
It certainly was an issue but maybe now that it's more stable it may no longer be. So let's roll with this to avoid this I forgot about this UX issue and it is an annoying one

TomasVotruba:
Thank you, I also struggled with this while using scoper in deps

PR #468, opened 2021-04-14
update symfony to latest v4.4, drop support for ^3.2 and allow ^5.2

smoench:
This is just a draft for trying out if there is anything left upgrading to symfony v5.

- [ ] Box needs to allow Symfony v5


smoench:
Thanks 😊

PR #482, opened 2021-05-18
Validate the prefix in the Configuration instead of the command

theofidry:
BC break: the prefix can no longer end with back-slashes


marekdedic:
Hi,
I've been messing a little bit with php-scoper again and this caught me by surprise - not only is it that the prefix cannot end with backslashes, it cannot **contain** backslashes... I, for example, am using the prefix `Sgdg\Vendor` - now this fails...

theofidry:
I see, I suppose it's fine if it contain backslashes

PR #488, opened 2021-05-30
Move creation of the config into a dedicated factory class

theofidry:


theofidry:
The tests will be failing since Box relies on the current API

PR #499, opened 2021-06-13
Fix scoping of typed properties

### LINKED ISSUE #498: "self" is scoped to "prefix\self"
| Question    | Answer
| PHP-Scoper version | PHP Scoper version 0.15.0 2021-05-10 21:02:42 UTC
| PHP version | PHP 8.0.7 (cli) (built: Jun  1 2021 18:43:05) ( NTS gcc x86_64 )
| Platform with version | Linux
| Github Repo | https://github.com/sebastianbergmann/phpunit/issues/4706

This ...
```php
final class Configuration
{
    private static ?self $instance = null;
```

... is scoped to

```php
final class Configuration
{
    private static ?\PHPUnit\self $instance = null;
```

... which results in
```
Fatal error: Cannot use 'PHPUnit\self' as class name as it is reserved in ...
```

theofidry:
Closes #498

### DIFF
--- a/specs/class/typed-properties.php
+++ b/specs/class/typed-properties.php
@@ -34,6 +34,10 @@ return [
 class A {
     public string $name;
     
+    public ?self $instance = null;
+    
+    public static ?self $staticInstance = null;
+    
     public ?B $foo;
 
     public function a() {}
@@ -46,6 +50,8 @@ namespace Humbug;
 class A
 {
     public string $name;
+    public ?self $instance = null;
+    public static ?self $staticInstance = null;
     public ?B $foo;
     public function a()
     {
@@ -96,6 +102,8 @@ namespace Foo;
 class A
 {
     public string $name;
+    public ?self $instance = null;
+    public static ?self $staticInstance = null;
     public ?B $foo;
     public function a()
     {
@@ -109,6 +117,8 @@ namespace Humbug\Foo;
 class A
 {
     public string $name;
+    public ?self $instance = null;
+    public static ?self $staticInstance = null;
     public ?B $foo;
     public function a()
     {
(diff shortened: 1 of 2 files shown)


theofidry:
@sebastianbergmann is this urgent or is it acceptable to fix it by either not typing it or using a patcher? There is several things I would like to finish for the next version (as there is several BCs I would prefer to keep all BCs in one version) so if it is I'll likely need to create another branch to do a release

sebastianbergmann:
> is this urgent

Not really. At the moment, this issue only affects a feature branch that will, hopefully soon, be merged into `master`. PHPUnit 10 will not be released any time soon.

> is it acceptable to fix it by either not typing it

Not sure what you mean here. Are you suggesting that I should not use property type declarations in PHPUnit's code?

> using a patcher

I have not worked with patchers before. What would this look like?

theofidry:
>Not sure what you mean here. Are you suggesting that I should not use property type declarations in PHPUnit's code?

At least for now; This only affect `self` as a property type AFAICS so it's not all property type declarations

>I have not worked with patchers before. What would this look like?

```php
// scoper.inc.php
return [
      // ...
            'patchers' => [
                static function (string $filePath, string $prefix, string $contents): string {
                    if ('src/path/to/Configuration.php' !== $filePath) {
                        return $contents;
                    }
      
                    return str_replace(
                        'private static ?\PHPUnit\self $instance = null;',
                        'private static ?self $instance = null;',
                        $contents,
                ),
        ],
];
```

or something along those lines

rotdrop:
This is quite annoying. I have a work-around like this in my `scoper.inc.php`:

```
diff --git a/scoper.inc.php b/scoper.inc.php
index 9723a9a7..c8c11947 100644
--- a/scoper.inc.php
+++ b/scoper.inc.php
@@ -106,5 +106,13 @@ return [
 
       return preg_replace('/namespace .+;/', '', $contents);
     },
+    // work around broken Scoper
+    function (string $filePath, string $prefix, string $contents): string {
+      return preg_replace(
+        '%[?]\S+\\\\self%',
+        '?self',
+        $content
+      );
+    },
   ],
 ];
```

In my case it affects `symfony/translation/MessageCatalogue.php`

Can you please this bug? Symfony is a widely used framework.

Kind thanks in advance!

PR #511, opened 2021-06-22
fix #384 scoper-autoload.php is dumped in the wrong vendor directory

### LINKED ISSUE #384: [Scoper] scoper-autoload.php is dumped in the wrong `vendor` directory
One of the libraries I am installing via composer has its own 'vendor' folder.

The library I am using is called [carbon-fields](https://github.com/htmlburger/carbon-fields/), which creates a folder 'vendor/htmlburger/carbon-fields/packages/vendor'.

The final output of php-scoper 'scoper-autoload.php' is always getting output to that folder, rather than the root 'vendor' folder.

This seems to be the offending line:
https://github.com/humbug/php-scoper/blob/1275a483b3cf3a06245bb0eab5763bde73a5a72e/src/Console/Command/AddPrefixCommand.php#L229

Perhaps the `usort` call just above this line should be comparing `return strlen($a) <=> strlen($b);` so that it get the shortest path, rather than the longest?

tw-360vier:
fixes: #384 scoper-autoload.php is dumped in the wrong vendor directory

Using composer-bin-plugin like suggested in the readme i also hit this bug.


theofidry:
Thanks for the PR; will check out shortly

theofidry:
Thank you @tw-360vier ; note however there won't be a release any time soon. There is a couple more BC changes I would like to bring (so I prefer to group them in one release)

Issue #534, opened 2021-08-10
Composer install for contribution fails on conflitcs


TomasVotruba:
Hi Theo,

I'm trying to contribute via fresh fork (https://github.com/TomasVotruba/php-scoper), but I fail to get bar composer install. 

I've tried various hackings like:

* change `"replace"` in `composer.json` to specific version
* use PHP 8, use PHP 7.4
* remove `composer.lock`
* run `composer install`
* run `composer update`
* run `composer install --no-interaction --no-progress --no-suggest --prefer-dist --ansi`

But nothing works for me.

<br>

## Steps to Reproduce

```
git clone https://github.com/TomasVotruba/php-scoper.git
cd php-scoper
composer install
```

<br>

```bash
  Problem 1
    - humbug/box is locked to version dev-master and an update of this package was not requested.
    - humbug/box dev-master requires humbug/php-scoper ^0.13.10 || ^0.14 -> found humbug/php-scoper[dev-master, 1.0.x-dev (alias of dev-master)] but it does not match the constraint. See https://getcomposer.org/dep-on-root for details and assistance.
  Problem 2
    - humbug/box dev-master requires humbug/php-scoper ^0.13.10 || ^0.14 -> found humbug/php-scoper[dev-master, 1.0.x-dev (alias of dev-master)] but it does not match the constraint. See https://getcomposer.org/dep-on-root for details and assistance.
    - humbug/box 3.x-dev is an alias of humbug/box dev-master and thus requires it to be installed too.
    - humbug/box is locked to version 3.x-dev and an update of this package was not requested.
```

Any ideas what I'm doing wrong?

TomasVotruba:
This is failing GitHub Action with the exact output: https://github.com/humbug/php-scoper/pull/535/checks?check_run_id=3290677049

![image](https://user-images.githubusercontent.com/924196/128865308-d4b612f7-4f87-486c-b9dd-8c9eca71cbc1.png)

(screenshot in case the GitHub Action result is deleted)

theofidry:
Hi,

Yes you need to do `source .composer-root-version` or alternatively set the `COMPOSER_ROOT_VERSION` env variable manually.

It might be worth adding to the contributing doc, but the regular install commands and others should be done via the Makefile which IIRC should already take care of that issue

TomasVotruba:
Thanks for answer. Can we make this somehow part of `composer.json` itself? I had similar bug with Simplify, but managed to remove it.

[...]

PR #535, opened 2021-08-10
[CI] Allow composer install without hustle

TomasVotruba:
Fixes https://github.com/humbug/php-scoper/issues/534


TomasVotruba:
The fix is based on reply from Seldaek (without link, not to bother him),
see https://github.com/composer/composer/issues/9368#issuecomment-718112361

composer.json
```diff
@@ -72,7 +71,7 @@
             "bin-links": false
         },
         "branch-alias": {
-            "dev-master": "1.0-dev"
+            "dev-master": "0.14-dev"
         }
```
TomasVotruba, line 75: This was the blocker actually. As the `humbug/box` package depends on 0.14, this version cannot be higher. It would be using different version than allowed.

.github/workflows/composer_install.yml
```diff
@@ -0,0 +1,34 @@
...
+            -   uses: actions/checkout@v2
+
+            -
+                name: Setup PHP
+                uses: shivammathur/setup-php@v2
+                with:
+                    php-version: ${{ matrix.package }}
+
+            -
+                uses: "ramsey/composer-install@v1"
```
TomasVotruba, line 34: See https://github.com/ramsey/composer-install#usage

This should simplify the other jobs, as this action handles caching :muscle: 

image

composer.json
```diff
@@ -62,17 +61,14 @@
 
     "config": {
         "bin-dir": "bin",
-        "sort-packages": true,
-        "platform": {
-            "php": "7.4.0"
-        }
```
TomasVotruba, line 68: This is not needed, as package should run on both PHP 7.4 and PHP 8.*

TomasVotruba:
Let's see how CI handles this...

.github/workflows/build.yml
```diff
@@ -83,10 +69,6 @@ jobs:
                     - 'e2e_032'
                 php: [ '7.4', '8.0' ]
                 tools: [ 'composer:v2' ]
-                include:
-                    -   php: '7.4'
-                        e2e: 'check-composer-root-version'
-                        tools: 'composer:v2'
```
TomasVotruba, line 89: This jobs is not needed anymore, handled by constant and correct alias

.github/workflows/build.yml
```diff
@@ -101,27 +83,7 @@ jobs:
...
-                        composer-
-
-            -   name: Set composer root version
-                run: |
-                    source .composer-root-version
-                    echo "COMPOSER_ROOT_VERSION=${COMPOSER_ROOT_VERSION}" >> $GITHUB_ENV
-
-            -   name: Install dependencies
-                run: composer install --no-interaction --no-progress --no-suggest --prefer-dist
+            -   uses: "ramsey/composer-install@v1"
```
TomasVotruba, line 86: This is basically equivalent of removed code above

TomasVotruba:
I'll split this PR into couple of smaller ones, so it's easier to follow what is fixed.
It will be also easier to "git blame" in the future in case of bugs.

theofidry:
See https://github.com/humbug/php-scoper/pull/537#issuecomment-1030684759

PR #536, opened 2021-08-11
[fix] bump php-parser to 4.12 and fix PHP 8.1 constants

TomasVotruba:
Split of https://github.com/humbug/php-scoper/pull/535

Fixes PHP 8.1 introduced native constants that should not be prefixed

Screenshot from 2021-08-11 08-41-24

PR #537, opened 2021-08-11
[CI] use COMPOSER_ROOT_VERSION and ramsey/composer-install

TomasVotruba:
Split of https://github.com/humbug/php-scoper/pull/535

Based on official composer docs for this issue: https://getcomposer.org/doc/articles/troubleshooting.md#dependencies-on-the-root-package

Also useful https://github.com/composer/composer/issues/9368#issuecomment-718112361

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -31,6 +31,9 @@ jobs:
                     tools: ${{ matrix.tools }}"
                     coverage: none
 
+            # set branch alias to allow install humbug/box that depends on 0.14
+            -   run: composer config extra.branch-alias.dev-master 0.14-dev
+
             -   name: Install dependencies
                 # see https://github.com/ramsey/composer-install
                 uses: "ramsey/composer-install@v1"
@@ -89,6 +92,9 @@ jobs:
                     tools: ${{ matrix.tools }}
                     coverage: pcov
 
+            # set branch alias to allow install humbug/box that depends on 0.14
+            -   run: composer config extra.branch-alias.dev-master 0.14-dev
+
             -   name: Install dependencies
                 # see https://github.com/ramsey/composer-install
                 uses: "ramsey/composer-install@v1"
@@ -116,6 +122,9 @@ jobs:
                     ini-values: 'phar.readonly=0'
                     tools: 'composer:v2'
                     coverage: 'none'
+
+            # set branch alias to allow install humbug/box that depends on 0.14
+            -   run: composer config extra.branch-alias.dev-master 0.14-dev
 
             -   name: Install dependencies
                 # see https://github.com/ramsey/composer-install
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,6 +7,11 @@ on:
     release:
         types: [ created ]
 
+env:
+    # see https://github.com/composer/composer/issues/9368#issuecomment-718112361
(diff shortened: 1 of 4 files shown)


TomasVotruba:
CI fails because waits on https://github.com/humbug/php-scoper/pull/536

.github/workflows/build.yml
```diff
@@ -25,27 +30,13 @@ jobs:
...
-                        composer-${{ runner.os }}-${{ matrix.php }}-
-                        composer-${{ runner.os }}-
-                        composer-
-
-            -   name: Set composer root version
-                run: |
-                    source .composer-root-version
-                    echo "COMPOSER_ROOT_VERSION=${COMPOSER_ROOT_VERSION}" >> $GITHUB_ENV
+            # set branch alias to allow install humbug/box that depends on 0.14
+            -   run: composer config extra.branch-alias.dev-master 0.14-dev
```
TomasVotruba, line 35: This is to keep current `composer.json` untouched. Though in the future would better to keep this line removed and `composer.json` + `humbug/box` synced. We use this approach in Symplify and Rector and it works well :+1:

theofidry:
@TomasVotruba thanks for the PR, they are not going unnoticed; unfortunately I'm a bit under water atm wrapping up stuff before going on vacation but hopefully I can take a look at them Sunday/Monday

TomasVotruba:
> @TomasVotruba thanks for the PR, they are not going unnoticed; unfortunately I'm a bit under water atm wrapping up stuff before going on vacation but hopefully I can take a look at them Sunday/Monday

I was a bit concerned about lack of feedback, so thanks for the update :+1: 

No worries, enjoy your vacation and relax :hugs:

.github/workflows/build.yml
theofidry: So do I understand correctly that you set `COMPOSER_ROOT_VERSION=dev-master` for forcing PHP-Scoper version to `dev-master` and then setting the alias to match whichever branch you want it to correspond to, i.e. `0.14-dev` here and as we bump versions we would need to bump this alias?

>Though in the future would better to keep this line removed and composer.json + humbug/box synced

What do you mean by this?

theofidry (commented):
Thanks for the PR @TomasVotruba!

Sorry for the big delay, a bit of a mix of hard work and holidays 😅 

I am not 100% sure of understanding the PR. I kinda get what's happening in the CI but how does it changes the issue locally?

.github/workflows/build.yml
TomasVotruba: I already forgot as it's been so long time ago for my contributions, but something like what you say.

By that I mean now humbug requires 2 different version, but box only one. That leads to collision. Ideally they should requires only one last mayor version.

TomasVotruba:
Hi, I don't remember anymore, I have dozens PRs a week :D. Probably with the alias it helped locally too.

theofidry:
Soooo after a long while I've checked more than once this PR and experimented various things. I am not really convinced this improves anything as you still need to load the composer root version AND additionally you need to set the branch alias. So the user does not really gain anything and I need to add additional tooling to ensure the branch alias version is correct.

I see however, that if both Box & PHP-Scoper were within the same repository, this would no longer be a problem (with your setup).

That being said, your PRs were still bringing various other improvements with #536 which I could somewhat port and augment, improving the overall project workflow. So this was definitely not lost work :)

PR #644, opened 2022-02-08
Add support for union and intersection types

### LINKED ISSUE #643: Scoper is not able to scope union return types
## Bug report

| Question    | Answer
| PHP-Scoper version | 0.16.2
| PHP version | 8.1
| Github Repo | - 

Given following class:
https://github.com/symfony/symfony/blob/6.1/src/Symfony/Component/Config/Definition/Builder/TreeBuilder.php#L35-L38

Actual
```php
public function getRootNode() : \Symfony\Component\Config\Definition\Builder\NodeDefinition|\Symfony\Component\Config\Definition\Builder\ArrayNodeDefinition
```

Expected:
```php
public function getRootNode() : \_PhpScoperae3dc8cfa6ce\Symfony\Component\Config\Definition\Builder\NodeDefinition|\_PhpScoperae3dc8cfa6ce\Symfony\Component\Config\Definition\Builder\ArrayNodeDefinition
```

veewee:
Fixes #643

### DIFF
--- a/specs/types/intersection.php
+++ b/specs/types/intersection.php
@@ -1,0 +1,145 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the humbug/php-scoper package.
+ *
+ * Copyright (c) 2017 Théo FIDRY <theo.fidry@gmail.com>,
+ *                    Pádraic Brady <padraic.brady@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+return [
+    'meta' => [
+        'title' => 'Union types',
+        // Default values. If not specified will be the one used
+        'prefix' => 'Humbug',
+
+        'expose-global-constants' => false,
+        'expose-global-classes' => false,
+        'expose-global-functions' => false,
+        'expose-namespaces' => [],
+        'expose-constants' => [],
+        'expose-classes' => [],
+        'expose-functions' => [],
+
+        'exclude-namespaces' => [],
+        'exclude-constants' => [],
+        'exclude-classes' => [],
+        'exclude-functions' => [],
+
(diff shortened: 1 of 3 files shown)


specs/union.php
```diff
@@ -0,0 +1,97 @@
...
+    
+    class X
+    {
+        public function foo(Y|Z $a, null|Y $b) : null|Y|Z
+        {
+        }
+    }
+    
+    PHP,
+    'Function casts' => <<<'PHP'
```
theofidry, line 61: I would suggest adding a blank line in-between

specs/union.php
```diff
@@ -0,0 +1,97 @@
...
+    }
+    
+    ----
+    <?php
+    
+    namespace Humbug;
+    
+    class X
+    {
+        public function foo(Y|Z $a, null|Y $b) : null|Y|Z
```
theofidry, line 55: might be worth adding `static` or `self` into the mix as well? Isn't there a nullable variant too?

If this is too bloated for a single method it's also find to have those cases within different functions:

```
        public function method0(Y|Z $a) : Y|Z
        public function method1(Y|null $a) : Y|null
// ...
```

theofidry (commented):
Thanks a lot! I'm really be happy if that's the only change necessary for adding union support 😄

specs/union.php
veewee: Added these cases as well and covered traits and interfaces as well.

veewee:
> Thanks a lot! I'm really be happy if that's the only change necessary for adding union support 😄

Yeah me too! Looks like the system is set up in a very flexible way. Good job!
I also took the liberty of adding support for intersection types. Or would you rather have a separate PR for that once this one is done?

We already created a version of grumphp that uses the union types with box, that seems to work ok.

theofidry:
Thank you @veewee

PR #651, opened 2022-02-10
Replace files-whitelist with exclude-files in docs

alexander-schranz:


theofidry:
Thank you @alexander-schranz

PR #669, opened 2022-03-23
Fix typo in option description

Kocal:


theofidry:
Thank you @Kocal! Sorry that it took a while

Issue #666, opened 2022-03-02
Exclude gettext function "__" seems to have side effects


hugofintecture:
## Bug report

| Question    | Answer
| ------------| ---------------
| PHP-Scoper version | 0.17.2 
| PHP version | 7.4.28
| Platform with version | WSL 2

When adding the "__" function (gettext for translations) in **exclude-functions** array, it seems to have unwanted-side effects.
This function is used for example in WordPress plugin in my case. This function should be outside the namespace to work.

In this case for example, during scoping, with the library `php-http/guzzle6-adapter` this lead to at least one error:

/vendor/php-http/guzzle6-adapter/src/Client.php

At line 65 there is this: 
`$handlerStack = new HandlerStack(\GuzzleHttp\choose_handler());`

While it should be (and it is ok when removing "__" from the exclude array):
` $handlerStack = new HandlerStack(\FintectureWC\GuzzleHttp\choose_handler());`

Maybe a side effect with the "_" character?

<details>
 <summary>scoper.inc.php</summary>
 
 ```php
<?php

declare(strict_types=1);

use Isolated\Symfony\Component\Finder\Finder;

// You can do your own things here, e.g. collecting symbols to expose dynamically
// or files to exclude.
// However beware that this file is executed by PHP-Scoper, hence if you are using
// the PHAR it will be loaded by the PHAR. So it is highly recommended to avoid
// to auto-load any code here: it can result in a conflict or even corrupt
// the PHP-Scoper analysis.

return [
    // The prefix configuration. If a non null value is be used, a random prefix
    // will be generated instead.
    //
    // For more see: https://github.com/humbug/php-scoper/blob/master/docs/configuration.md#prefix
    'prefix' => 'FintectureWC',

    // By default when running php-scoper add-prefix, it will prefix all relevant code found in the current working
    // directory. You can however define which files should be scoped by defining a collection of Finders in the
    // following configuration key.
    //
    // This configuration entry is completely ignored when using Box.
    //
    // For more see: https://github.com/humbug/php-scoper/blob/master/docs/configuration.md#finders-and-paths
    'finders' => [
        Finder::create()->files()->in('src'),
        Finder::create()
            ->files()
            ->ignoreVCS(true)
            ->notName('/LICENSE|.*\\.md|.*\\.dist|Makefile|composer\\.json|composer\\.lock/')
            ->exclude([
                'doc',
                'test',
                'test_old',
                'tests',
                'Tests',
                'vendor-bin',
                'examples'
[...]

Issue #597, opened 2022-01-25
Restrict possible delimiters for regexes


theofidry:
In PHP in practice any character can be used for regexes. In the context of PHP-Scoper, i.e. matching symbols fully qualified names against regexes, various delimiters are causing issues. For example `\` which makes `\Foo\A` a perfectly valid case or `Bar\BW` for which `B` works out.

For this reason I think it's simpler to restrict regexes to special characters (i.e. non unicode letters) and exclude `\` as well.

PR #671, opened 2022-04-18
Rejects _ as a valid regex delimiter, fixes issue #666.

bobab12:
This is one way to fix issue #666 which is related to #597.

Alternatively it could be implemented by rejecting empty regex. However, there appears to be a test to explicitly accept empty regex so it made sense to simply reject `_` as a delimiter altogether.


theofidry:
Merged, thanks a lot @bobab12, sorry that it took so long

PR #706, opened 2022-11-04
Fix scoping of internal symbol declaration

### LINKED ISSUE #682: Make it easier to scope global functions
Unfortunately this is not a problem that will go away any time soon as long as big frameworks such as Laravel & Symfony do not try to change, and worse continue even for new code, the practice of:

- introducing new functions in the global scope
- not using fully-qualified function names or import statement for functions, i.e. relying on the PHP function autoload fallback mechanism

An idea to tackle this problem would be to expose yet another setting, e.g. `user-global-functions = list<string|regex>` which would have the following effects:

- it considers this function as registered and declared within the scope code. This means that it _will_ be namespaced. For example:

Original code:

```php
<?php // vendor/symfony/deprecation-contracts/functions.php

if (!function_exists('trigger_deprecation')) {
   function trigger_deprecation() { ...}
}
```

(currently) Scoped code:

```php
<?php // vendor/symfony/deprecation-contracts/functions.php

namespace Scoped;

if (!function_exists('trigger_deprecation')) { // <- this declaration is now incorrect, the new setting would help fix that
   function trigger_deprecation() { ...}
}
```

- the function name would not be registered as internal and would kick in if the call is not FQ, e.g. using `trigger_deprecation()` in the code (without the leading `\` or use statement), would be resolved to `Scoped\trigger_deprecation`.

theofidry:
Closes #682


src/PhpParser/NodeVisitor/NamespaceStmt/FunctionIdentifierRecorder.php
```diff
@@ -31,7 +31,7 @@
 use PhpParser\NodeVisitorAbstract;
 
 /**
- * Records the user functions registered in the global namespace which have been whitelisted and whitelisted functions.
+ * Records the exposed functions used or the???
```
theofidry, line 34: TODO

PR #811, opened 2022-12-30
Check if function `humbug_phpscoper_expose_class` exists

### LINKED ISSUE #809: Fatal error: Cannot redeclare humbug_phpscoper_expose_class()
## Bug report

| Question    | Answer
| PHP-Scoper version | 0.18.1 |
| PHP version | 8.1 |
| Platform with version | MacOS 13 |
| Github Repo | https://github.com/leoloso/PoP |

I'm using PHP-Scoper to scope 2 WordPress plugins. When I install both of them on the same site, I get the following error:

```
Fatal error: Cannot redeclare humbug_phpscoper_expose_class() (previously declared in /app/wordpress/wp-content/plugins/graphql-api/vendor/scoper-autoload.php:23) in /app/wordpress/wp-content/plugins/graphql-api-pro/vendor/scoper-autoload.php on line 27
```

The reason is clear: both plugins define function `humbug_phpscoper_expose_class` in the global space.

Maybe we can append `if(!function_exists('humbug_phpscoper_expose_class')) { ...` when [declaring the function](https://github.com/humbug/php-scoper/blob/f7bd92f2459f1d9a643313f6d324476b0e23e087/src/Autoload/ScoperAutoloadGenerator.php#L44-L50)?

For reference, this is [one of my `scoper.inc.php`](https://github.com/leoloso/PoP/blob/1a7cda04a30e4426fd34516788f8155c7f16849f/ci/scoping/scoper-graphql-api.inc.php).

leoloso:
Fixes #809

@theofidry I can't run the tests locally, as `composer update` is failing (am I doing something wrong?):

```
  Problem 1
    - humbug/box[4.0.0, ..., 4.2.0] require composer-plugin-api ^2.2 -> found composer-plugin-api[2.1.0] but it does not match the constraint.
    - Root composer.json requires humbug/box ^4.0 -> satisfiable by humbug/box[4.0.0, ..., 4.2.0].
```

But I can see the tests are passing here on GitHub


theofidry:
>I can't run the tests locally, as composer update is failing (am I doing something wrong?)

If you do `make phpunit` it should work. The issue is that you need to source `.composer-root-version` since there is a circular dependency (in dev) with Box.

theofidry (approved):
👍 The patch looks good to me thanks!

I would like to capture this into a dedicated test for the future so I'll hold off a bit for the merge. Hopefully I should be able to figure it out soon :)

theofidry:
Thanks a lot @leoloso!

PR #846, opened 2023-09-29
Fixed scoper-autoload.php being incompatible with PHP 5

### LINKED ISSUE #842: scoper-autoload.php doesn't support PHP 5.6
## Bug report

| Question    | Answer
| PHP-Scoper version | 0.18.3
| PHP version | 5.6
| Platform with version | all
| Github Repo | [- ](https://github.com/skaut/skaut-google-drive-gallery/)

Hi,
I am using PHP-scoper for packaging a WordPress plugin, so I don't really get much control over which PHP versions it will be running with. Recently I got reports of ` syntax error, unexpected ‘:’, expecting ‘{‘` in the `scoper-autoload.php` file and was surprised to find that it contains return type annotations, possibly as an artifact from the build process. I tried to look for info about supported runtime php versions in the project, but couldn't find any

I assume composer minimum version is the version of PHP php-scoper runs on, but the output would respect the source code it processes...

marekdedic:
Closes #842

### DIFF
--- a/src/Autoload/ScoperAutoloadGenerator.php
+++ b/src/Autoload/ScoperAutoloadGenerator.php
@@ -43,7 +43,7 @@ final class ScoperAutoloadGenerator
 
     private const EXPOSE_CLASS_DECLARATION = <<<'PHP'
         if (!function_exists('humbug_phpscoper_expose_class')) {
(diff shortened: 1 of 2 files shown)


theofidry:
Thank you @marekdedic. I cannot say when this will be released I'll try to have a look ASAP

Issue #810, opened 2022-12-28
Error on Windows


AnirudhSingal91:
## Bug report

| Question    | Answer
| Getting error while running the php-scoper add-prefix command  

I came across php-scoper because I was experiencing  dependency hell in my custom wordpress plugin. I just want to prefix one package(guzzlehttp). I found implementing php-scoper to be extremely hard and i'm probably doing it all wrong. 

I get the following error while running the php-scoper add-prefix command

In Filesystem.php line 95:
                                                                                                                                                                                                                                                    
Failed to create "C:/Users/singa/source/repos/WIP/woocommerce-xero-integration/woocommerce-xero-integration/woocommerce-xero-integration/dependencies/vendorC:\Users\singa\source\repos\WIP\woocommerce-xero-integration\woocommerce-xero-integration\woocommerce-xero-integration\vendor\guzzlehttp\guzzle\src": mkdir(): No such file or directory  


| ---------------
| PHP-Scoper version | x.y.z <!-- (`php-scoper.phar -V` or `composer show | grep humbug/php-scoper`) -->0.18
| PHP version | x.y.z <!-- (`php -v`) -->8.1
| Platform with version | e.g. Ubuntu/Windows/MacOS Windows
| Github Repo | - <!-- (if public) -->


<!--
Replace this comment with your issue description. Please complete the table above
with the correct information when relevant and include if relevant:
- The steps to reproduce your issue
- The exact command run
- Your configuration file

Also ensure the issue lies with the isolated code and not with shipping the code in a PHAR. Please
refer to the documentation (#Recommendations) for more details.

For general support, please use the #humbug Slack channel: https://symfony.com/slack-invite.
-->

<details>
 <summary>scoper.inc.php</summary>
 
 ```php
 <?php
 
 declare(strict_types=1);
 
 use Isolated\Symfony\Component\Finder\Finder;
 
 return [
 ];

 ```
</details>

<details>
 <summary>Output</summary>
 
 ```bash
 $ command
 > output
 ```
</details>

theofidry:
@AnirudhSingal91 sorry to hear.

Unfortunately this library is not tested under Windows. I have no Windows machine myself so it was never practical to add support for it and no one really stepped up to help out there. A few issues have occasionally been fixed for it, so maybe it works, maybe not.

I unfortunately cannot help out much there without any sort of reproducer.

You might also want to checkout #303 which talks to length about the WP use case.

AnirudhSingal91:
[...]

PR #862, opened 2023-10-20
Normalize paths to avoid issue on Windows filesystems

liqueurdetoile:
Related to https://github.com/humbug/php-scoper/issues/810

Hi,

I was facing the same issue after updating one of my package and doing local testings on a Windows mahcine. I've checked this fix locally and it's doing fine. I've used normalize instead of canonicalize as it's a bit faster and I don't think that weird paths like `to/../to/` are expected.

Tests workflow is green on my fork but I'm facing unexpected failure on E2E tests. I can't see why this commit could brain out PHAR generation. Let's see how this this PR performs 😸 

Feel free to close it if it's not welcome.

### DIFF
--- a/src/Console/ConsoleScoper.php
+++ b/src/Console/ConsoleScoper.php
@@ -180,9 +180,9 @@ final class ConsoleScoper
         Assert::notNull($commonDirectoryPath);
 
         $mapFiles = static fn (array $inputFileTuple) => [
-            $inputFileTuple[0],
+            Path::normalize($inputFileTuple[0]),
             $inputFileTuple[1],
-            $outputDir.str_replace($commonDirectoryPath, '', $inputFileTuple[0]),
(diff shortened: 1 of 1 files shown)


theofidry:
Thank you @liqueurdetoile!

theofidry:
>Tests workflow is green on my fork but I'm facing unexpected failure on E2E tests. I can't see why this commit could brain out PHAR generation. Let's see how this this PR performs 😸

It could be they just don't work on Windows, since I don't have a windows machine I never tried

liqueurdetoile:
You're welcome 😺

liqueurdetoile:
Also thanks @EusebiuOprinoiu  for the findout in code.

For E2E issue, it was on my forked repository. I presumed there was some missing artifacts because I've not enabled all the workflows, maybe ?

theofidry:
Ha that can be too indeed

Issue #683, opened 2022-09-26
Scoped folder name of package not what expected?


alessandrotesoro:
## Bug report

| Question    | Answer
| ------------| ---------------
| PHP-Scoper version | 0.17.5
| PHP version | 8
| Platform with version | MacOS

Hi there,

In a project I'm working on, we have 2 packages with the same vendor name, example "my-company". When scoping the packages, they're both scoped, however the folder structure of one of them is completely broken. 

In the configuration below: the package with name "first-package" is correctly scoped under `dependencies/first-package` - the 2nd package however, for some reason, it's using the entire project structure and becomes `dependencies/project-name-here/vendor/my-company/second-package` but what I'd need it to be is `dependencies/second-package`.

Any idea what's causing this?

Note: the 1st package is symlinked locally. The 2nd isn't.

<details>
 <summary>scoper.inc.php</summary>
 
 ```php
 <?php

declare( strict_types=1 );

use Isolated\Symfony\Component\Finder\Finder;

return [
    // The prefix configuration. If a non null value will be used, a random prefix will be generated.
    'prefix' => 'My_Company\\Plugin\\Test_Plugin\\Dependencies',

    /**
     * By default when running php-scoper add-prefix, it will prefix all relevant code found in the current working
     * directory. You can however define which files should be scoped by defining a collection of Finders in the
     * following configuration key.
     *
     * For more see: https://github.com/humbug/php-scoper#finders-and-paths.
     */
    'finders'                    => [
        Finder::create()->
            files()->
            ignoreVCS( true )->
            ignoreDotFiles( true )->
            notName( '/LICENSE|.*\\.md|.*\\.dist|Makefile|composer\\.(json|lock)/' )->
            exclude(
                [
                    'doc',
                    'test',
                    'build',
                    'test_old',
                    'tests',
                    'Tests',
                    'vendor-bin',
                    'wp-coding-standards',
                    'squizlabs',
                    'phpcompatibility',
                    'dealerdirect',
                    'bin',
                    'vendor'
                ]
            )->
            in( [
                'vendor/my-company/first-package',
                'vendor/my-company/second-package'
            ] )->
        name( [ '*.php' ] ),
    ],

    'patchers'                   => [
        function ( string $file_path, string $prefix, string $contents ): string {
            // Change the contents here.

            return str_replace(
                'Symfony\\\\',
                sprintf( '%s\\\\Symfony\\\\', addslashes( $prefix ) ),
                $contents
            );
        },
    ],
];

[...]

PR #865, opened 2023-10-31
Changed output subpath logic

LuigiPulcini:
Concerning #683, the proposed changes are aimed at guaranteeing the subpath in the output directory is returning the correct values even when libraries are symlinked from a local location.

Since a Symfony Finder component returns files in an associative array where keys are the path of files always relative to the current project even when symlinks are followed, those keys can be effectively used to determine the subpath of the files in the output directory. The advantage is that there will not be any difference when the local library is replaced with the production one.

Please consider that, unlike what we discussed in #683, this first approach replaces the logic behind the current implementation. If this works, we could consider letting the two different strategies coexist through the appropriate configuration.

### DIFF
--- a/src/Configuration/ConfigurationFactory.php
+++ b/src/Configuration/ConfigurationFactory.php
@@ -390,7 +390,7 @@ final class ConfigurationFactory
     {
         $filesWithContents = [];
 
-        foreach ($files as $filePathOrFileInfo) {
+        foreach ($files as $fileKey => $filePathOrFileInfo) {
             $filePath = $filePathOrFileInfo instanceof SplFileInfo
                 ? $filePathOrFileInfo->getRealPath()
                 : realpath($filePathOrFileInfo);
@@ -413,7 +413,7 @@ final class ConfigurationFactory
                 );
             }
 
-            $filesWithContents[$filePath] = [$filePath, file_get_contents($filePath)];
+            $filesWithContents[$fileKey] = [$filePath, file_get_contents($filePath)];
         }
 
         return $filesWithContents;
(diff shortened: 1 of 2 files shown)


theofidry:
I'm happy it looks like small changes and could only get a quick look at the CI. My guess is that it has to do with the `exclude-files`, which are currently taken as realpath and I assume need to be adjusted in a similar fashion

LuigiPulcini:
Thanks again for taking action so promptly on this PR.

I initially thought that excluded files required special treatment but then I realized that both arrays are built using the same `retrieveFilesWithContents()` method [here](https://github.com/humbug/php-scoper/blob/main/src/Configuration/ConfigurationFactory.php#L98) and [here](https://github.com/humbug/php-scoper/blob/main/src/Configuration/ConfigurationFactory.php#L105), so I assumed that changing that method would suffice to make both arrays compatible with the proposed change.

Admittedly, I haven't tested the changes in a scenario where files are excluded so I would take a closer look at this detail and make the appropriate adjustments.

LuigiPulcini:
The latest commit attempts to fix the issue with files listed in the `exclude-files` property. Since we have to assume that the list is not an associative array, `$index` should be replaced with `$fileKey`, determined as `$file` after replacing the leading `$dirPath.DIRECTORY_SEPARATOR` with an empty string.

theofidry:
I think there is some work left to do here:

- Add a case with symlinks to show that it works
- `ConsoleScoper::getFiles()` appears to look for the common path to map the files and in one case (`make e2e_027`) this appears to not work (which kinda makes sense, but then this common path logic needs to be adapted).

LuigiPulcini:
I haven't addressed the two points you mentioned last year. I only needed to update my branch with the latest main branch. I will see if I can work on those two points later this month.

PR #880, opened 2023-11-02
add docs for excluding third-party symbols

calvinalkan:


docs/configuration.md
```diff
@@ -267,6 +268,46 @@ It is recommended to use excluded symbols only to complement the
 [PhpStorm's stubs][phpstorm-stubs] shipped with PHP-Scoper.
 
 
+### Excluding Third-Party Symbols
```
theofidry, line 271: I think if you phrase it like so, it's is very confusing as redundant with the previous doc section. And I actually realise there is a section "How to deal with unknown third-party symbols" in "Further Reading".

I think the two existing sections are still relevant, after all what your package is is not so much documenting the (already documented) feature but promoting a canonical way to deal with wordpress plugins.

So instead, maybe a `Wordpress Support` under "Further Reading" would be more appropriate?

calvinalkan: Yes, that sounds good.

PR #881, opened 2023-11-02
test: Add an e2e test for PHPUnit

### LINKED ISSUE #640: Add better PHPUnit support
See https://gist.github.com/sebastianbergmann/9eba10f9e61c576950df8bcd37712992#file-phar-md.

It would be interesting to have PHPUnit, as we have Infection, part of the e2e test suite.

theofidry:
Closes #640.


sebastianbergmann:
I think you should only use the [PHAR-specific tests](https://github.com/sebastianbergmann/phpunit/blob/main/.github/workflows/ci.yml#L228).

PR #897, opened 2023-11-04
doc: Fix the broken link even more

szepeviktor:
There was a missing dash.

image

From #896


szepeviktor:
Here are two tools for Markdown

1. markdownlint-cli2 for formatting
2. lychee for checking URL-s

https://github.com/szepeviktor/byte-level-care/blob/master/.github/workflows/markdown.yml#L32

theofidry:
Thanks! will look into it :)

PR #899, opened 2023-11-05
Update .gitattributes

szepeviktor:
Remove non-existent files, add new development files.

Test command: `git archive HEAD | tar -t --exclude='*/*'`

One more to add: `composer.lock` - the lock file is not used in packages.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -3,24 +3,27 @@
 *.php text eol=lf
 
 # Ignore all test and documentation with "export-ignore".
+/.composer-root-version          export-ignore
 /.gitattributes                  export-ignore
 /.gitignore                      export-ignore
-/.styleci.yml                    export-ignore
-/.travis.yml                     export-ignore
 /.github/                        export-ignore
+/.php-cs-fixer.dist.php          export-ignore
+/.scrutinizer.yml                export-ignore
 /_specs/                         export-ignore
 /bin/infection.sh                export-ignore
 /box.json.dist                   export-ignore
 /build/                          export-ignore
 /composer-root-version-checker/  export-ignore
 /docs/                           export-ignore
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -3,24 +3,27 @@
...
 /build/                          export-ignore
 /composer-root-version-checker/  export-ignore
 /docs/                           export-ignore
+/Dockerfile                      export-ignore
 /infection.json.dist             export-ignore
 /fixtures/                       export-ignore
+/.makefile/                      export-ignore
 /Makefile                        export-ignore
-/phpstan.neo.dist                export-ignore
+/phpstan.neon.dist               export-ignore
```
szepeviktor, line 23: A typo

theofidry: Thanks!

szepeviktor: Glad to contribute.

theofidry:
Thank you @szepeviktor

szepeviktor:
Actually fixing all these is a solution for **today**.
The Integrity workflow keeps all these fixed at every commit.
https://github.com/szepeviktor/byte-level-care/blob/master/.github/workflows/reusable-integrity.yml#L39
.. maybe this is only valid in my golden cage

theofidry:
I can see a few being helpful, not sure but a few others, but maybe I'll try adding those in the future :)

szepeviktor:
For example the one that looks for gitignored files in the index.
```
$ git ls-files --cached --ignored --exclude-standard
bin/infection.sh
bin/php-scoper
build/.gitkeep
fixtures/set028-symfony/expected-output
fixtures/set030/expected-output
```
Investigating ... 🕵🏻‍♂️

PR #900, opened 2023-11-05
build: Remove non-existent ignored files

szepeviktor:
@theofidry Here is another fruit 🍇 of the Integrity workflow.

From https://github.com/humbug/php-scoper/pull/899#issuecomment-1793703505

Now the output is empty.
```shell
git ls-files --cached --ignored --exclude-standard
```


theofidry:
Thanks!

PR #901, opened 2023-11-05
Make make help colorful and simple

szepeviktor:
Colorful and simple.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,7 @@ BLACKFIRE = blackfire
 
 .PHONY: help
 help:
-	@printf '\033[33mUsage:\033[0m\n  make TARGET\n\n\033[32m#\n# Commands\n#---------------------------------------------------------------------------\033[0m\n'
+	@printf '\033[33mUsage:\033[0m\n  make TARGET\n\n\033[32m#\n# Commands\n#---------------------------------------------------------------------------\033[0m\n\n'
 	@sed -n -e 's|^\(.\+:\) ##\(.\+\)$$|\o033[33m\1\o033[0m \2|p' $(MAKEFILE_LIST)
 
 
--- a/Makefile
+++ b/Makefile
@@ -38,8 +38,8 @@ BLACKFIRE = blackfire
 
 .PHONY: help
 help:
-	@echo "\033[33mUsage:\033[0m\n  make TARGET\n\n\033[32m#\n# Commands\n#---------------------------------------------------------------------------\033[0m\n"
-	@fgrep -h "##" $(MAKEFILE_LIST) | fgrep -v fgrep | sed -e 's/\\$$//' | sed -e 's/##//' | awk 'BEGIN {FS = ":"}; {printf "\033[33m%s:\033[0m%s\n", $$1, $$2}'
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -38,8 +38,8 @@ BLACKFIRE = blackfire
 
 .PHONY: help
 help:
-	@echo "\033[33mUsage:\033[0m\n  make TARGET\n\n\033[32m#\n# Commands\n#---------------------------------------------------------------------------\033[0m\n"
-	@fgrep -h "##" $(MAKEFILE_LIST) | fgrep -v fgrep | sed -e 's/\\$$//' | sed -e 's/##//' | awk 'BEGIN {FS = ":"}; {printf "\033[33m%s:\033[0m%s\n", $$1, $$2}'
+	@echo -e "\033[33mUsage:\033[0m\n  make TARGET\n\n\033[32m#\n# Commands\n#---------------------------------------------------------------------------\033[0m\n"
```
szepeviktor, line 41: The `-e` option was missing.

Makefile
```diff
@@ -38,8 +38,8 @@ BLACKFIRE = blackfire
...
+	@sed -n -e 's|^\(.\+:\) ##\(.\+\)$$|\o033[33m\1 \o033[0m\2|p' $(MAKEFILE_LIST)
```
szepeviktor, line 42: One `sed` does it all.

szepeviktor:
Why does  fail?
On my machine
```
$ make help
\033[33mUsage:\033[0m\n  make TARGET\n\n\033[32m#\n# Commands\n#---------------------------------------------------------------------------\033[0m\n
```

theofidry:
The result after running `make help` before/after the changes of your PR for me:

Screenshot 2023-11-05 at 22 42 55

Screenshot 2023-11-05 at 22 43 04

szepeviktor:
Is your shell Bash?

theofidry:
zsh*

szepeviktor:
I see.
`echo` is not portable. https://unix.stackexchange.com/a/700679

theofidry:
I indeed use this form in another project https://github.com/theofidry/makefile/blob/main/Makefile#L42.

Note I also didn't use sed, because it behaves different on OSX which is super annoying. And actually I think it's the culprit for the failure. The CI works fine as it's ubuntu, but locally I get the same as the "after" screenshot from before.

szepeviktor:
> Note I also didn't use sed, because it behaves different on OSX

Yes. GNU sed has an option: `sed --posix` to disable GNU extensions.
Indeed it does not work with that option!

theofidry:
TIL

szepeviktor:
The "one or more" `+` sign is the culprit.

I give up. Sorry.

theofidry:
So I think if you revert just the `@sed` change it's still a mergeable change :) I would be happy to simplify the current grep command but I admit I also didn't care enough to try to figure a way to do so (it really didn't look simple and the sort of issues I could spend hours on)

theofidry:
>I give up. Sorry.

No worries, it's a nasty one

szepeviktor:
> it really didn't look simple

Compatibility is a hot topic!
I keep investigating in my secret kitchen ...

PR #902, opened 2023-11-05
Mark Container properties nullable

szepeviktor:
... as all of them begin as `null`.


theofidry:
>... as all of them begin as null.

AFAICS they begin as "unset". So not sure what the nullability would bring here? If you check the get method you will never get a `null` instance

szepeviktor:
You are right.
This was a bad idea.

PR #952, opened 2024-03-08
fix: Do not exclude all source code in the default template

### LINKED ISSUE #946: All files are added to the `excluded-files` in a default scoper.inc.php
HI,

I initialized php-scoper on a new project (with the default `scoper.inc.php` ) and noticed that after the `add-prefix` command is finished, every single file which was prefixed in the process is reverted to it's initial state (not prefixed).
I noticed that in the default `scoper.inc.php`, every file inside the directory of .inc file is added to the list of `excluded-files` recursively.

theofidry:
Closes #946.


theofidry:
Thank you for the report @sui-drmz

PR #1024, opened 2024-04-29
Fixed scoper-autoload.php PHP 5.6 incompatibility

### LINKED ISSUE #1023: scoper-autoload.php doesn't support PHP 5.6 (again)
## Bug report

| Question    | Answer
| PHP-Scoper version | 0.18.7
| PHP version | 5.6
| Platform with version | all
| Github Repo | -

This is a followup to #842. I found another thing in `scoper-autoload.php` that is incompatible with PHP 5.6 - the `??` operator. PR incoming.

See the error here: https://github.com/skaut/skaut-google-drive-gallery/actions/runs/8883783876/job/24391429247?pr=2632

marekdedic:
Closes #1023


src/Autoload/ScoperAutoloadGenerator.php
```diff
@@ -103,7 +103,7 @@ public function dump(): string
                 namespace {
                     \$loader = (static function () {
                         // Backup the autoloaded Composer files
-                        \$existingComposerAutoloadFiles = \$GLOBALS['__composer_autoload_files'] ?? [];
+                        \$existingComposerAutoloadFiles = isset(\$GLOBALS['__composer_autoload_files']) ? \$GLOBALS['__composer_autoload_files'] : [];
```
theofidry, line 106: wouldn't the `trait_exists()` call L47 fail too?

theofidry (commented):
I thought the Composer autoloader now requires PHP 7.2 minimum?

src/Autoload/ScoperAutoloadGenerator.php
marekdedic: `trait_exists` is present on PHP >= 5.4
https://www.php.net/manual/en/function.trait-exists.php

But yeah, an automated check of the generated file would be great, I only encountered this while trying to set up PHP 5.6 compatibility testing, which was a royal pain itself... I don't really know hoe to do that, maybe run it through PHPCS with PHPCompatibility checks?

marekdedic:
> I thought the Composer autoloader now requires PHP 7.2 minimum?

I don't think so - the autoload script definitely checks if it's on PHP 5.6 or older, so I suppose it should support very old PHP...? You can't run composer on such old PHP, but that's a different thing from running the generated autoloader...

src/Autoload/ScoperAutoloadGenerator.php
theofidry: Guess the easiest would be via a docker image, doing an e2e test like there is in the `.makefile/e2e`. But hardly a requirement

theofidry:
Thank you @marekdedic, I can't promise this sticks around without any automated test for it but it doesn't hurt to have

PR #1035, opened 2024-05-18
fix: Handle symbols dependencies

### LINKED ISSUE #1010: Fatal error: Uncaught Error: Interface 'Stringable' not found in
## Bug report

| Question    | Answer
| PHP-Scoper version | master branch dure to this possiblee fix 
| PHP version | 8.2

After running php-scoper it seems that some interfaces are no scoped.

See the source of file /vendor/symfony/polyfill-php80/PhpToken.php
```
namespace Symfony\Polyfill\Php80;

class PhpToken implements \Stringable
{
....
}
```

and file vendor/symfony/polyfill-php80/Resources/stubs/Stringable.php
```
if (\PHP_VERSION_ID < 80000) {
    interface Stringable
    {
        /**
         * @return string
         */
        public function __toString();
    }
}
```

are scoped like

```
namespace ReinosMapsScoped\Symfony\Polyfill\Php80;

class PhpToken implements \Stringable <!-- NOTICE THID
{
...
}
```

```
if (\PHP_VERSION_ID < 80000) {
    interface Stringable
    {
        /**
         * @return string
         */
        public function __toString();
    }
    \class_alias('ReinosMapsScoped\\Stringable', 'Stringable', \false);
}
```

The issue seems that `class PhpToken implements \Stringable` is not scoped to `class PhpToken implements \ ReinosMapsScoped\Stringable`

Now it throws an PHP error `Fatal error: Uncaught Error: Interface 'Stringable' not found in.....` and with adding manually the scoped namespace it fixed the error.

Is this a bug in PHPscoper or a configuration issue? 🤔

theofidry:
Attempt to fix #1010.


fixtures/set041-exposed-symbols-hierarchy/stubs/StringeableLike.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+if (\PHP_VERSION_ID < PHP_INT_MIN) {
+    interface StringableLike
```
reinos, line 4: Should this not be `StringeableLike`

theofidry: Ha indeed; but did not solve the problem 🤔

PR #1051, opened 2024-06-16
fix: Configure the printer to preserve formatting

theofidry:


kkmuffme:
Also in case the new printer still does not print unmodified (excluded-namespace,...) files as they were originally, I think it would be a simple improvement to not use the phpparser printer for that then, but just print the original input as is.

PR #1061, opened 2024-07-20
Fix function with multiple namespaces

VincentLanglet:
Hi @theofidry, I discovered a bug with the php-scoper.

Before the fix, I get
```
namespace Acme\Emca {
    if (!function_exists('Acme\Emca\foo')) { function Emca\foo() { return \Humbug\Acme\Emca\foo(...func_get_args()); } }
}
```
which is crashing.

PR #1065, opened 2024-08-15
Fix tests

VincentLanglet:
I think there is a change in the behavior of some dependency which remove this useless parenthesis when parsing the file.

```
($file instanceof \SplFileInfo) ? $file : new \SplFileInfo($file);
```
is transformed to
```
$file instanceof \SplFileInfo ? $file : new \SplFileInfo($file);
```

Since the parenthesis is useless, it seems better to remove it from the both input and output IMHO.

### DIFF
--- a/tests/AutoReview/GAE2ETest.php
+++ b/tests/AutoReview/GAE2ETest.php
@@ -31,7 +31,7 @@ class GAE2ETest extends TestCase
 
     public function test_github_actions_executes_all_the_e2e_tests(): void
     {
-        $expected = array_values(array_diff(E2ECollector::getE2ENames(), self::IGNORED_E2E_TESTS));
+        $expected = array_diff(E2ECollector::getE2ENames(), self::IGNORED_E2E_TESTS);
         $actual = GAE2ECollector::getExecutedE2ETests();
 
         self::assertEqualsCanonicalizing($expected, $actual);
--- a/tests/AutoReview/GAE2ETest.php
+++ b/tests/AutoReview/GAE2ETest.php
@@ -31,7 +31,7 @@ class GAE2ETest extends TestCase
 
     public function test_github_actions_executes_all_the_e2e_tests(): void
     {
-        $expected = array_diff(E2ECollector::getE2ENames(), self::IGNORED_E2E_TESTS);
+        $expected = array_values(array_diff(E2ECollector::getE2ENames(), self::IGNORED_E2E_TESTS));
         $actual = GAE2ECollector::getExecutedE2ETests();
 
         self::assertEqualsCanonicalizing($expected, $actual);
--- a/composer.lock
+++ b/composer.lock
@@ -207,16 +207,16 @@
         },
         {
             "name": "nikic/php-parser",
-            "version": "v5.0.2",
+            "version": "v5.1.0",
             "source": {
                 "type": "git",
                 "url": "https://github.com/nikic/PHP-Parser.git",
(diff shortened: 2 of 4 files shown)


VincentLanglet:
I don't understand the failure https://github.com/humbug/php-scoper/actions/runs/10401466964/job/28820621264?pr=1065 ; it might be something random.

For the removal of the parenthesis it comes from the next version of nikic/php-parser
https://github.com/nikic/PHP-Parser/releases/tag/v5.1.0

So I had to update the version in the composer.lock in order to use the same for all the tests @theofidry

VincentLanglet:
Can you approve again the CI @theofidry, to see if the PR is ok or need more fixes ?

tests/AutoReview/GAE2ETest.php
```diff
@@ -31,7 +31,7 @@ class GAE2ETest extends TestCase
 
     public function test_github_actions_executes_all_the_e2e_tests(): void
     {
-        $expected = array_diff(E2ECollector::getE2ENames(), self::IGNORED_E2E_TESTS);
+        $expected = array_values(array_diff(E2ECollector::getE2ENames(), self::IGNORED_E2E_TESTS));
```
theofidry, line 34: I don't understand this change: isn't the `assertEqualsCanonicalizing()` supposed to not care about the keys difference?

theofidry (commented):
the unit test failures looks related to some changes in the jetbrains stubs. I don't know if that was intentional of them or not, and it's worth noting that those stubs are about the IDE, they are not meant to provide an exhaustive list of symbols (e.g. there's plenty of old unmaintained extensions that are not documented).

In any case such changes go to the Enriched reflector. Usually I try to keep the reflector as light as possible, but add more tests for checking against regressions. If it's not obvious I may be able to take a lookt this week

VincentLanglet:
> the unit test failures looks related to some changes in the jetbrains stubs. I don't know if that was intentional of them or not, and it's worth noting that those stubs are about the IDE, they are not meant to provide an exhaustive list of symbols (e.g. there's plenty of old unmaintained extensions that are not documented).
> 
> In any case such changes go to the Enriched reflector. Usually I try to keep the reflector as light as possible, but add more tests for checking against regressions. If it's not obvious I may be able to take a lookt this week

ares_gethostbyname was removed in https://github.com/JetBrains/phpstorm-stubs/pull/1627 because it's considered as outdated.

I assume it's the same for 
uv_read2_start
uv_handle_type
uv_ares_init_options

Do you know these method ? Is the PR on phpstorm-stubs wrong ?

theofidry:
I don't know them, but it's what I meant by "those stubs are about the IDE, they are not meant to provide an exhaustive list of symbols" when talking of JetBrains stubs. So whilst we can use it to leverage a big number of stubs, that list needs to be complemented.

The missing symbols can be added to https://github.com/humbug/php-scoper/blob/main/src/Symbol/Reflector.php

VincentLanglet:
> The missing symbols can be added to [`main`/src/Symbol/Reflector.php](https://github.com/humbug/php-scoper/blob/main/src/Symbol/Reflector.php?rgh-link-date=2024-08-30T13%3A10%3A15Z)

Done @theofidry

theofidry:
Thank you @VincentLanglet!

VincentLanglet:
> Thank you @VincentLanglet!

Thanks you too.
I rebased https://github.com/humbug/php-scoper/pull/1061

PR #1090, opened 2024-12-02
feat: Add support for PHP 8.4 property hooks

theofidry:


tests/Symbol/Reflector/PhpStormStubsReflectorTest.php
```diff
@@ -98,6 +98,20 @@ public static function provideClasses(): iterable
             'MongoInsertBatch',
             'MongoDeleteBatch',
         );
+
+        yield from self::createDataSetForInternalSymbols(
+            'HP 8.4 new class-like (added or modified)',
```
staabm, line 103: typo?

```suggestion
            'PHP 8.4 new class-like (added or modified)',
```

theofidry: yes!

theofidry: FYI I'll probably won't have time to look into before new year, kinda under water at work and then I'll benefit from being more off for the festivities

PR #1100, opened 2025-01-02
Closes #1099

### LINKED ISSUE #1099: Dependency `jetbrains/phpstorm-stubs` needs to be updated to `v2024.3` for PHP 8.4 compatibility
Quoting myself from https://github.com/humbug/php-scoper/issues/1092#issuecomment-2567472851:

> `use PropertyHookType;` in `Generator.php` is [...] transformed into `use PHPUnitPHAR\PropertyHookType;`.

The above seems to happen because PHP-Scoper 0.18.15 comes with version `v2024.2` of the `jetbrains/phpstorm-stubs` which does not know about classes such as `PropertyHookType` that were introduced in PHP 8.4. The first version of `jetbrains/phpstorm-stubs` to support PHP 8.4 is `v2024.3`.

sebastianbergmann:


theofidry:
Thank you @sebastianbergmann

PR #1207, opened 2026-05-05
Do not prefix scalar strings with known internal class names

ondrejmirtes:
I'm trying to solve a problem. Right now in 0.18.19, php-scoper prefixes code such as:

```php
$subclasses['PDO\Mysql'] = 'PDO\Mysql';

return new ObjectType('Filter\FilterFailedException')
```

into:

```php
$subclasses['PDO\Mysql'] = '_PHPStan_6ef81ee5d\PDO\Mysql';

return new ObjectType('_PHPStan_6ef81ee5d\Filter\FilterFailedException')
```

Since these are known internal classes, it should not be happening.


theofidry (commented):
Does it make sense to solve that problem? Yes.

I actually thought this was already taken care of 🤦 .

Is the presented fix correct? I have to check. I'm having doubts about excluded vs. exposed atm. The line changed over time, and I need to re-evaluate that.

But I believe the direction is correct, so if that solves it I may defer that assessment later in favour of getting this out ASAP

